PR #101, opened 2025-11-24
Long running task monitor

PiyushMahore:


dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java
```diff
@@ -109,14 +108,35 @@ record CacheKey(String peerIdentity, long id){}
 
     private final Path tempDir;
 
+    private final ScheduledExecutorService scheduledExecutorService;
+
+    private final Duration maxStatementLifetime;
```
tanejagagan: This should be named queryTimeout

PR #316, opened 2026-04-11
disable TLS certificate validation in HttpArrowProducer

PiyushMahore:
Uses a trust-all X509TrustManager and disables hostname verification so self-signed and internal certificates are accepted without needing them added to the JVM truststore. Affects all callers (dazzleduck-sql-logback and dazzleduck-sql-micrometer) since they both delegate to HttpArrowProducer.

Adds TrustAllCertificateTest which spins up a local HttpsServer with a keytool-generated self-signed cert and asserts that ingestion succeeds over HTTPS without an SSLHandshakeException.


dazzleduck-sql-common/src/main/java/io/dazzleduck/sql/common/SslUtils.java
```diff
@@ -0,0 +1,46 @@
...
+            return ctx;
+        } catch (NoSuchAlgorithmException | KeyManagementException e) {
+            throw new RuntimeException("Failed to create trust-all SSLContext", e);
+        }
+    }
+
+    /**
+     * Creates an HttpClient that accepts any certificate and skips hostname verification.
+     */
+    public static HttpClient trustAllHttpClient() {
```
admin-one211, line 83: Improve this implementation based on. environment variable. If DD_TRUST_SELF_SIGNED_CERTS then return trutAllHttpClient else return default client
Add this to documentation

dazzleduck-web (approved):

PR #362, opened 2026-07-20
Eliminate unused LEFT JOINs when inlining a view

gtaneja-sentrinox:
## What

Adds `Transformations.pruneUnusedLeftJoins(conn, outerSqlAst, viewBodyAst)` — a pure AST-in/AST-out optimization for the common **fact/dimension view** pattern.

Given an outer query whose `FROM` is a single view reference, plus the parsed view body, it:
1. **Inlines** the view as a subquery in place of the reference,
2. **Prunes** the view's `SELECT` list to the columns the outer query actually uses (projection pushdown),
3. **Drops** `LEFT JOIN`s to dimension tables whose columns are then referenced nowhere (iterated to a fixpoint).

### Example

```sql
-- view fv: SELECT f.*, a.a_name, b.b_name
--          FROM f LEFT JOIN a ON f.a_id=a.a_id LEFT JOIN b ON f.b_id=b.b_id
SELECT f_col, a_name FROM fv WHERE f_col > 10;
-- becomes:
SELECT f_col, a_name
FROM (SELECT f.*, a.a_name FROM f LEFT JOIN a ON f.a_id=a.a_id) AS fv
WHERE f_col > 10;          -- LEFT JOIN b eliminated, b_name projection dropped
```

## Design & scope (v1)

Full design in `LEFT_JOIN_PRUNING_SPEC.md`.

- **LEFT-join only**; the null-supplying right arm must be a base table with an equi-join condition.
- **Join-key uniqueness is trusted** (standard star schema) — no verification. This is the single documented correctness assumption: a non-unique dimension could change row counts. Called out loudly in the spec.
- **Optimization, never a semantic change.** On any uncertainty the method returns the **same instance** it was given (callers detect no-op by identity).

### Correctness guards

- Any STAR in the outer query — bare `*` **or qualified `fv.*`** — bails (both expand to view columns that can't be enumerated at the AST level).
- Projection pushdown is skipped under `DISTINCT` / `GROUP BY` / grouping sets / modifiers (where dropping a select entry would change row count or shift positional refs like `GROUP BY 1`); **join elimination still applies**.
- Multi-part refs (`s.field`) are treated conservatively so struct/nested columns are never pruned away.
- `USING(...)` / non-`REGULAR` joins, non-single-base-table outer `FROM` → bail.

### Performance

The transform is pure in-memory AST work: **~7 µs** (2 joins) to **~13 µs** (8 joins). One reference-count walk per fixpoint round, not per candidate join. The surrounding SQL⇄AST round-trips (`json_serialize_sql`/`json_deserialize_sql`, ~70–100 µs each) dominate — negligible against any real query, and reusable if the caller already parses to AST.

## Tests

- `PruneUnusedLeftJoinsTest` — elimination, transitive/fixpoint, DISTINCT, GROUP BY, LIMIT, nested `STRUCT` column, and bail-outs (qualified star, bare star, USING, multi-table FROM). Equivalence verified with `TestUtils.isEqual` against the real view.
- `PruneUnusedLeftJoinsDuckLakeTest` — fact table is a DuckLake table with the hidden `rowid` meta column projected by the view; elimination preserves `rowid` and the unmatched fact row.

No production call site yet — this lands the transformation + spec + tests. Wiring into the query path is a follow-up.

### DIFF
--- a/LEFT_JOIN_PRUNING_SPEC.md
+++ b/LEFT_JOIN_PRUNING_SPEC.md
@@ -161,6 +161,10 @@ no-op by reference identity. Explicit bail-outs:
 - `USING(...)` join (null `condition`, populated `using_columns`).
 - `right` arm not a base table; non-equi `ON` condition.
 - More than one `BASE_TABLE` in the outer `FROM`.
+- **Outer `FROM` resolves to a WITH-declared CTE** — a local CTE shadows any
+  catalog view of the same name (`WITH fv AS (…) SELECT … FROM fv`), so the
+  reference is the CTE, not the view; inlining the view body there would change
+  results.
 
 Additional conservative rules:
 
--- a/LEFT_JOIN_PRUNING_SPEC.md
+++ b/LEFT_JOIN_PRUNING_SPEC.md
@@ -1,0 +1,248 @@
+# Unused LEFT JOIN Elimination — Design Specification
+
+## Overview
+
+Given an **outer query that selects from a view**, rewrite it so that
+`LEFT JOIN`s inside the view whose columns the outer query never references are
+removed — producing a semantically equivalent but cheaper query.
+
+Two chained optimizations run: **projection pushdown** (the outer query prunes
+the view's `SELECT` list) followed by **join elimination** (a table that then
+contributes no referenced column is dropped).
+
+**Status:** specification + scaffolded failing tests. Implementation is a no-op
+stub (`Transformations.pruneUnusedLeftJoins`) pending the work described here.
+
+---
+
+## Worked Example
+
+```sql
+-- view
+CREATE VIEW fv AS
+SELECT f.*, a.a_name, b.b_name
+FROM   f
+LEFT JOIN a ON f.a_id = a.a_id
+LEFT JOIN b ON f.b_id = b.b_id;
+
+-- outer query (input)
+SELECT f_col, a_name FROM fv WHERE f_col > 10;
+
+-- output
+SELECT f_col, a_name
+FROM (
+  SELECT f.*, a.a_name          -- b_name projection dropped (projection pushdown)
+  FROM   f
+  LEFT JOIN a ON f.a_id = a.a_id -- LEFT JOIN b eliminated (join elimination)
+) AS fv
+WHERE f_col > 10;
+```
+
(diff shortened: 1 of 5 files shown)


gtaneja-sentrinox:
## Review + CTE coverage (commit f3dff19)

Reviewed the transform end-to-end — the design is sound: pure AST-in/AST-out, same-instance no-op on any uncertainty, LEFT-preservation verified against the b-unmatched fact row, and the trusted join-key-uniqueness assumption is documented loudly. Bail-outs (stars, `USING`, non-equi ON, multi-table FROM, DISTINCT/GROUP BY projection skip) all look correct and are well tested.

While adding CTE tests I found one **correctness gap**, now fixed on this branch.

### Bug: a CTE shadowing the view name changed results

A local CTE shadows any catalog view of the same name, so:

```sql
WITH fv AS (SELECT 42 AS f_col, 'zz' AS a_name, 'yy' AS b_name)
SELECT f_col FROM fv
```

Here `fv` is the **CTE**, not the view. The method identified the outer `FROM` as a base table named `fv`, matched it to the passed view body, and inlined that body in place of the CTE reference — rewriting the query to return the fact rows (`5,15,25`) instead of the CTE's `42`. That is exactly the semantics change the transform promises never to make (a caller that fetches a view `fv` from the catalog cannot see that the query locally rebinds `fv` as a CTE).

### Fix

Added a `referencesOuterCte` guard: if the single outer `FROM` base-table name matches a `WITH`-declared CTE in the same query, bail (return the input instance). SQL scoping guarantees the CTE always wins, so this can only ever be the wrong target. Documented in the spec's bail-out list.

### Tests added (`PruneUnusedLeftJoinsTest`)

- `cteInOuterQuery_viewStillPrunedAndCteUsageCounted` — outer `FROM` is the view with an auxiliary CTE consumed by a scalar subquery; pruning still drops the unused `b` join, and `collectUsage` correctly descends into `cte_map`.
- `dimColumnUsedOnlyInCte_joinKept` — a dimension column referenced **only** inside a CTE keeps its join (proves CTE usage is counted, not just the main SELECT/WHERE).
- `cteShadowsViewName_returnedUnchanged` — the regression guard for the bug above (no-op / same instance).

All 16 `PruneUnusedLeftJoinsTest` + 3 `PruneUnusedLeftJoinsDuckLakeTest` pass (JDK 21).

PR #365, opened 2026-07-22
feat(commons): prune view LEFT JOINs when the view is referenced inside a CTE

gtaneja-sentrinox:
## What

Follow-up to #362. Adds a scope-aware overload:

```
pruneUnusedLeftJoins(outerSqlAst, viewName, viewBodyAst)
```

that prunes a fact/dimension view's unused `LEFT JOIN`s wherever the view is referenced — at the outer top-level `FROM` **and/or inside any number of CTE bodies** — locating references by (optionally qualified) name.

### Example (previously a no-op, now optimized)

```sql
-- view fv: SELECT f.*, a.a_name, b.b_name
--          FROM f LEFT JOIN a ON f.a_id=a.a_id LEFT JOIN b ON f.b_id=b.b_id
WITH a AS (SELECT a_name FROM fv) SELECT * FROM a;
-- becomes (LEFT JOIN b eliminated, b_name dropped):
WITH a AS (SELECT a_name FROM (SELECT f.*, a.a_name FROM f LEFT JOIN a ON f.a_id=a.a_id) AS fv)
SELECT * FROM a;
```

## Design

The v1 transform was hardwired to one scope — "the top-level SELECT whose FROM is the view." This makes it scope-aware. **Every eligible reference is pruned independently within its own scope** — inlining is local and semantics-preserving, so any subset of references may be optimized:

- **Top-level** — usage collected from the outer SELECT scope *excluding* CTE bodies (a CTE cannot reference the outer FROM's columns), so sibling CTE usage doesn't pollute top-level pruning.
- **CTE-nested** — each CTE body is its own scope: usage and the STAR check are evaluated there, so an outer `SELECT *` *over the CTE* (which expands to CTE columns, not view columns) doesn't block the optimization.
- A scope containing a STAR (view columns not enumerable) is **skipped individually**; other references still prune.

**Qualified names**: `viewName` may be `fv`, `s2.fv`, or `cat.s2.fv`. Matching is level-exact in both directions — an unqualified name never matches a qualified reference (may be a different, same-named view), and a qualified name never matches an unqualified reference (search-path resolution is invisible at the AST level). Qualified references can never resolve to a CTE, so they are immune to CTE shadowing.

## Correctness / bail-outs (same-instance no-op)

Everything the v1 method guarded, plus:
- **Unqualified** view name rebound by a CTE anywhere in the outer `WITH` (sibling or outer shadow, incl. nested WITH inside a CTE body) — the reference resolves to the CTE, not the view.
- Qualification mismatch in either direction (see above).

Still pure AST-in/AST-out; any structural surprise degrades to the no-op.

## Review findings fixed during the PR (see comments)

- `isViewRef` originally matched by bare `table_name` — a CTE body referencing a same-named view in another schema (`s2.fv`) got the *wrong* view body inlined. Fixed with level-exact qualification matching.
- `referencesOuterCte` matched CTEs by bare name even for qualified refs, blocking legitimate top-level pruning of `s3.fv` when a CTE named `fv` existed. Qualified refs now never match CTEs.

## Scope (intentional limits)

Not yet: a CTE body doing `SELECT *` from the view while the outer selects a subset (needs projection-through-a-star across scopes — that scope is skipped, others still prune); view references nested deeper than a scope's direct FROM (inside joins/subqueries); quoted identifiers containing dots.

## Tests

`PruneUnusedLeftJoinsTest` grew 17 → 35, covering: the CTE-wrapping example (elimination + row preservation + equivalence vs the real view), per-scope pruning of two CTE references, top-level + CTE pruned independently (a different join survives in each scope), star-scope skipped while siblings prune, dim column used only in the CTE body, nested/sibling CTE shadowing, schema-qualified same-named view regression, qualified-name matching (CTE-nested, top-level, vs-unqualified no-op, immune to bare-name CTE shadow), aliased refs, and dual reference (CTE + outer join). All 35 + 3 DuckLake tests pass (JDK 21).

### DIFF
--- a/LEFT_JOIN_PRUNING_SPEC.md
+++ b/LEFT_JOIN_PRUNING_SPEC.md
@@ -166,6 +166,27 @@ no-op by reference identity. Explicit bail-outs:
   reference is the CTE, not the view; inlining the view body there would change
   results.
 
+### Scope-aware variant — view referenced inside a CTE
+
+`pruneUnusedLeftJoins(outerSqlAst, viewName, viewBodyAst)` extends the above to the
+case where the view is referenced inside a **CTE body** rather than at the outer
+top-level `FROM`, e.g. `WITH a AS (SELECT a_name FROM fv) SELECT * FROM a`. Passing
+`viewName` lets a reference be located by name. Two shapes are handled; everything
+else is the same-instance no-op:
+
+- **Top-level** — outer `FROM` is the view: delegates to the two-arg method, behaviour
+  unchanged.
+- **CTE-nested** — the view is in the `FROM` of exactly one CTE body. Column usage and
+  the STAR bail are evaluated **within that CTE body scope**, so an outer `SELECT *`
+  *over the CTE* (which expands to CTE columns, not view columns) does not block the
+  optimization. The pruned view body is inlined in place of the reference inside the CTE
+  body; the CTE's own `SELECT` list is untouched.
+
+Additional bail-outs for this variant: the view name is rebound by a CTE anywhere in the
+outer `WITH` (sibling/outer shadow); a bare/qualified STAR **within the CTE body** over
+the view; the view referenced by **more than one** CTE body (single reference only in
+this increment).
+
 Additional conservative rules:
 
 - **Projection pushdown is skipped** (join elimination still applies) when the
--- a/LEFT_JOIN_PRUNING_SPEC.md
+++ b/LEFT_JOIN_PRUNING_SPEC.md
@@ -182,11 +182,17 @@ else is the same-instance no-op:
   optimization. The pruned view body is inlined in place of the reference inside the CTE
   body; the CTE's own `SELECT` list is untouched.
 
-Additional bail-outs for this variant: the view name is rebound by a CTE anywhere in the
-outer `WITH` (sibling/outer shadow); a bare/qualified STAR **within the CTE body** over
-the view; the view referenced by **more than one** CTE body (single reference only in
-this increment); a **schema- or catalog-qualified** reference (`s2.fv`) — `viewName` is
-unqualified, and a qualified reference may be a *different*, same-named view.
+`viewName` may be **qualified** (`s2.fv`, `cat.s2.fv`); qualification must match the
+reference at the same level. An unqualified name matches only unqualified references —
+a qualified reference (`s2.fv`) may be a *different*, same-named view. A qualified name
+matches only identically-qualified references — an unqualified reference resolves via
+the search path, which is invisible at the AST level. Qualified references can never
+resolve to a CTE, so qualified names skip the CTE-shadow bail below.
+
+Additional bail-outs for this variant: an **unqualified** view name rebound by a CTE
+anywhere in the outer `WITH` (sibling/outer shadow); a bare/qualified STAR **within the
+CTE body** over the view; the view referenced by **more than one** CTE body (single
+reference only in this increment).
 
 Additional conservative rules:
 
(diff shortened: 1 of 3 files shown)


gtaneja-sentrinox:
## Review (commit a7b7a77)

Reviewed the scope-aware CTE increment end-to-end and probed the edge cases the original tests didn't cover.

### ✅ Verified sound
- **Nested-WITH shadowing** — `WITH a AS (WITH fv AS (…) SELECT … FROM fv) SELECT * FROM a` correctly bails: `findSoleCteReferencingView` checks the CTE body's own `cte_map` via `referencesOuterCte(body, from)`.
- **Dual reference** — view in a CTE body *and* the outer FROM join: only the CTE body is rewritten; the outer `fv` keeps pointing at the real view; results verified equivalent. Correct because inlining is local and semantics-preserving.
- **Aliased reference** (`FROM fv x`) prunes fine — `effectiveId` picks up the alias for the subquery wrapper.
- Top-level delegation to the v1 path is behavior-identical; the fail-safe (same-instance no-op on any surprise) is preserved throughout.

### 🐛 Bug found & fixed: schema-qualified same-named view got the wrong body inlined
`isViewRef` compared only `table_name`. A CTE body referencing a **different, same-named view in another schema**:

```sql
WITH a AS (SELECT a_name FROM s2.fv) SELECT * FROM a
```

matched the unqualified `viewName = "fv"` and had the **default-schema** view body inlined in place of `s2.fv` — silently swapping the view (probe test reproduced this: `'other'` replaced by real fact data). Since `viewName` is unqualified in this increment, a schema- or catalog-qualified reference now never matches (bails).

### Tests added
- `schemaQualifiedSameNamedView_returnedUnchanged` — regression for the bug
- `nestedWithInsideCteBodyShadowsView_returnedUnchanged`
- `viewInCteAndOuterFromJoin_cteBodyPruned_outerRefUntouched`
- `aliasedViewRefInCteBody_pruned`

All 28 `PruneUnusedLeftJoinsTest` + 3 `PruneUnusedLeftJoinsDuckLakeTest` pass (JDK 21). Spec bail-out list updated.

gtaneja-sentrinox:
## Review of the qualified-viewName commit (9fadcc7) → fix in a798ee2

Reviewed the qualified-name matching end-to-end and probed its interaction with the earlier shadow-bail logic.

### ✅ Verified sound
- **Level-exact matching** is the right call in both directions: unqualified name ↛ qualified ref (may be a different, same-named view — the bug fixed in a7b7a77), and qualified name ↛ unqualified ref (search-path resolution is invisible at the AST level).
- `QualifiedName.parse` degrades safely: >3 parts → unmatchable → the standard no-op. Dotted quoted identifiers are documented as unsupported.
- CTE-nested path correctly skips the shadow bail for qualified names (`WITH fv AS (…), a AS (SELECT a_name FROM s3.fv) …` prunes).

### 🔧 Inconsistency found & fixed: top-level delegation still applied the bare-name shadow bail
```sql
WITH fv AS (SELECT 1 AS t)
SELECT f_col, a_name FROM s3.fv WHERE f_col > (SELECT t FROM fv)
```
with `viewName = "s3.fv"` was a **no-op**: the top-level path delegates to the two-arg method, whose `referencesOuterCte` matched the CTE named `fv` by bare `table_name` — although a qualified reference always resolves to the catalog and can never hit a CTE. Missed optimization only (fails safe), but inconsistent with the CTE-nested path.

**Fix:** `referencesOuterCte` now returns false for schema-/catalog-qualified refs — one principled change that corrects both the v1 API and the delegation.

### Test-quality note
My first version of the probe asserted only `!sql.contains("b_name")` — vacuously green on a no-op, since the *outer* query text never mentions `b_name`. Strengthened to `assertNotSame` + join count. Existing negative-signal assertions elsewhere in the suite pair with `countJoins`/equivalence checks, so they're fine.

All 33 `PruneUnusedLeftJoinsTest` + 3 DuckLake tests pass (JDK 21).

### Residual notes (no action)
- Name matching is case-sensitive; DuckDB unquoted identifiers are case-insensitive, so a case-mismatched `viewName` degrades to the no-op (conservative).
- Malformed names like `"fv."` parse as unqualified `fv` (Java `split` drops trailing empties) — garbage-in, still safe.

PR #373, opened 2026-08-06
fix(ingestion): reliability fixes for BulkIngestQueue and OTel collector error handling

gtaneja-sentrinox:
## Summary

Fixes four reliability issues in the bulk-ingestion path found during performance/reliability testing, plus observability for a fifth:

1. **Failed writes permanently inflated `pendingWrite` (critical, self-bricking).** `pendingWrite()` was `acceptedBytes - totalWrite`, and only successful writes updated `totalWrite` — failed bytes counted as pending forever, so after enough storage failures every `add()` was rejected at `maxPendingWrite` until a process restart. Failed buckets are now accounted in dedicated `failedWrite{Bytes,Batches,Buckets}` counters and subtracted from `pendingWrite()` / `getPendingBatches()`. This also corrects the inflated `calculateRetryAfterSeconds()` estimates. Write failures are now logged (previously swallowed silently).

2. **Producer sequence advanced before a successful write.** Retrying a batch whose write failed was rejected as `OutOfSequenceBatch`, turning recoverable failures into permanent data loss. On write failure, each affected producer's sequence entry is rolled back below its smallest failed id (under the queue lock, before the futures are failed), so the exact failed batches can be resubmitted. Duplicate protection for successfully written batches is unchanged.

3. **Flush trigger scheduling leak.** Every size-triggered flush spawned an additional never-terminating `triggerWriteIfRequired` chain; under sustained ingestion thousands of chains accumulated, all contending for the queue lock (~10% observed throughput degradation). A single-flight `triggerScheduled` guard keeps exactly one chain alive: the constructor starts it and each run schedules its successor.

4. **OTel collector returned backpressure as `UNKNOWN` with an empty message.** `PendingWriteExceededException` is now mapped to `RESOURCE_EXHAUSTED` with the original message and a `RetryInfo` status detail carrying the queue's retry-after estimate (honored by OTLP exporters for throttling); existing `StatusRuntimeException`s pass through; other failures map to `INTERNAL` with the root-cause message. This matches the mapping the Flight (`ErrorHandling`) and HTTP (`IngestionService`, 429 + `Retry-After`) surfaces already had.

5. **Producer-id LRU evictions were silent.** Evicting one of the 10,000 tracked producer ids silently disabled duplicate/ordering protection for that producer. Evictions are now logged with the producer id and counted via a `producer_id_evictions` metric; the limit is documented.

New per-queue metrics are exposed through `Stats`, the OTel writer counters (`bytes_failed`, `batches_failed`, `write_failures`, `producer_id_evictions`), and the Flight recorder counter map.

## Test plan

- New regression tests in `BulkIngestQueueTest`:
  - failed write corrects `pendingWrite()` and the queue accepts new batches that phantom bytes would have rejected
  - a failed batch can be retried with the same producer batch id; duplicates of written batches still rejected; multi-batch failed buckets fully retriable
  - 50 size-triggered flushes add zero scheduled trigger chains, and the single chain still flushes via the delay path
  - producer-id eviction is counted and the evicted producer demonstrably loses duplicate protection
- New tests in `OtelServiceBaseTest` for the gRPC status mapping (RESOURCE_EXHAUSTED + RetryInfo round-trip, INTERNAL with message, StatusRuntimeException pass-through)
- Full `dazzleduck-sql-commons` suite green (383 tests); `dazzleduck-sql-otel-collector` suite green; `dazzleduck-sql-flight` compiles with the new counters

### DIFF
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/BulkIngestQueue.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/BulkIngestQueue.java
@@ -454,7 +454,15 @@ public abstract class BulkIngestQueue<T, R> implements BulkIngestQueueInterface<
             return;
         }
         triggerScheduled = true;
-        executorService.schedule(this::triggerWriteIfRequired, delayMillis, TimeUnit.MILLISECONDS);
+        try {
+            executorService.schedule(this::triggerWriteIfRequired, delayMillis, TimeUnit.MILLISECONDS);
+        } catch (RuntimeException e) {
+            // e.g. RejectedExecutionException from an externally shut-down executor. Leave the
+            // flag clear so a later add()/flush can restart the chain; a stuck-true flag would
+            // silently disable all time-based flushes while the queue still accepts batches.
+            triggerScheduled = false;
+            throw e;
+        }
     }
 
     private synchronized void submitWriteTask() {
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/BulkIngestQueue.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/BulkIngestQueue.java
@@ -14,6 +14,7 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


gtaneja-sentrinox:
Added two follow-up commits:

**`0dabbeaf` — repaired `OtelCollectorBenchmark`.** It was silently failing 100% of requests on main (missing `x-dd-ingestion-queue` JWT claim, arrow extension never loaded because only `Main` runs the startup script, output directories never created). With the fixes it produces real numbers — local baseline: ~162k records/sec at 16 clients × 10k-record batches, with p50 pinned at `maxDelay` (1s), i.e. flush-cadence bound, not writer bound, on a fast local disk.

**`0a6de050` — DuckLake two-phase commit-profile scenario** in `BulkIngestQueueLoadTest`, as the baseline for the parallel-commit-lanes enhancement. It models the commit as a parallelizable data phase (DuckDB COPY, internally multi-threaded) followed by a globally serialized catalog phase (DuckLake single-writer metadata update, via an injectable shared lock). With 150ms + 50ms latencies the single commit thread saturates exactly like the production plateau: 2.15 MB/s measured vs 2.50 MB/s ceiling at 96% writer utilization. It also prints the Amdahl bound for lane scaling ((data+catalog)/catalog = 4.0× for this profile), since the catalog phase stays serialized regardless of lane count.

gtaneja-sentrinox:
**`c820e44f` — per-phase commit instrumentation in `ParquetIngestionQueue`.**

The write is now timed as two phases: the **data phase** (DuckDB COPY to Parquet — internally multi-threaded, overlappable across future commit lanes) and the **post-ingestion phase** (e.g. the DuckLake `ducklake_add_data_files` catalog transaction — single-writer, serialized regardless of lane count). Exposed as cumulative counters on both metrics surfaces: `dazzleduck.otel.writer.data_phase_ms` / `post_ingest_phase_ms` (OTel collector) and `data_phase_ms` / `post_ingest_phase_ms` (Flight recorder), plus a per-write debug log line.

The point: the production ratio `post / (data + post)` read off a dashboard directly gives the Amdahl bound for the parallel-commit-lanes enhancement — max speedup = `(data + post) / post` — so the lane design (and whether it's worth building) can be decided from real numbers rather than estimates.

gtaneja-sentrinox:
## Self-review results (`75b753ca`)

**Critical (fixed):** `ParquetIngestionQueue.write()` swallowed every exception — it completed the futures itself and returned normally, so `BulkIngestQueue.processWriteQueue` treated failures as successes. On the real Parquet/DuckLake path this meant the headline fixes of this PR never engaged: failed bytes were counted into `totalWrite` as written, the new `failed_*` counters stayed zero, and producer sequences never rolled back (retries still rejected as `OutOfSequenceBatch`). The mock queues throw from `write()`, which is why all earlier tests passed. `write()` now rethrows and the base class owns failure handling in the correct order (account → roll back sequences → complete futures). New regression test drives a failed write through the real class and asserts accounting + same-batch-id retry.

**Medium (fixed):** `scheduleTrigger` left the single-flight flag stuck `true` if `executorService.schedule` threw (e.g. `RejectedExecutionException`), which would permanently disable time-based flushes while the queue kept accepting batches. The flag is now cleared on scheduling failure.

**Minor (fixed):** `toStatusError`'s cause-chain walk is now depth-bounded (pathological cause cycles).

**Noted, not changed:**
- Success-path ordering: futures complete inside `write()` *before* the base class accumulates `totalWrite`, so `pendingWrite()` can be momentarily stale after a client observes success. Harmless (backpressure errs briefly on the conservative side); a clean fix would be `write()` returning the result and the base class completing futures — a wider refactor than this PR should carry.
- Partial-failure semantics: a `write()` that completes some futures before throwing counts the whole bucket's bytes as failed. Pre-existing ambiguity; metrics err on the side of overcounting failures.

Full suites green after the fixes: commons 378, flight 147, otel-collector 50.

PR #374, opened 2026-08-14
feat(ingestion): per-group MIN-timestamp watermark in the DuckLake post-ingestion transaction

gtaneja-sentrinox:
## What

Adds an optional **watermark append** to the DuckLake post-ingestion path: for a configured queue, one `MIN(<timestamp>)` row per group is inserted into a watermark table **in the same transaction** as the `ducklake_add_data_files` registration. File registration and watermark rows commit or roll back together, so a consumer can never observe registered files without their watermark, or watermark rows for files that were rolled back.

## How

Configured per queue mapping through the existing `additional_parameters` map:

```hocon
additional_parameters {
    watermark_table            = "ingest_watermark"  # same catalog/schema as the target table
    watermark_timestamp_column = "timestamp"         # required when watermark_table is set
    watermark_group_columns    = "county,state"      # optional; empty = one global row per batch
}
```

**Watermark rows are computed at write time, not read back from the written files.** `ParquetIngestionQueue` aggregates the same pre-COPY relation the output Parquet is written from (local data, transformation applied), carries the rows on `IngestionResult.watermarkRows`, and `DuckLakePostIngestionTask` appends them as a plain `INSERT ... VALUES` alongside the registration calls. Consequences:

- no re-read of just-written files (no second S3 round-trip)
- no coupling to the written files' schema (heterogeneous batches register exactly as before)
- partition columns group as real typed columns (they are hive-path-only in the files)
- rows whose MIN would be NULL (empty batch, all-NULL timestamps) are excluded by construction

**Config is validated at load time** (`QueueIdToTableMapping` construction): partial specs, blank values, typo'd `watermark_*` keys, and malformed group lists fail startup with a message naming the key, instead of failing per batch after output is written. Flattened HOCON-list syntax for `watermark_group_columns` is tolerated. A new `WatermarkSpec` record owns parsing, validation, and both SQL shapes; identifiers are quoted via the existing `HeaderUtils.quoteIdentifier`, and string arguments to `ducklake_add_data_files` (including file paths) are escaped.

Known limitation (documented): queues registered via the dynamic SQLite provider don't carry `additional_parameters`, so watermarks require statically configured mappings.

Documented in the otel-collector `reference.conf` queue-mapping block.

## Testing

Full `dazzleduck-sql-commons` suite under JDK 21: **383 tests, 0 failures, 0 errors** — including the Arrow-based `ParquetIngestionQueueTest` / `DuckLakeIngestionIntegrationTest`, which exercise the refactored write path end to end.

New coverage:
- `WatermarkSpecTest` — parsing (absent/full/HOCON-list/partial/blank/typo'd), load-time validation via `QueueIdToTableMapping`, per-group MIN with NULL-timestamp exclusion, empty/all-NULL suppression in global mode, INSERT rendering (quoted identifiers, escaped values)
- `DuckLakeWatermarkPostIngestionTest` — precomputed rows appended in the same transaction (min-across-files case), **rollback atomicity** (failing watermark insert rolls back file registration), empty-rows and no-config paths unchanged

### DIFF
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
@@ -179,6 +179,13 @@ public class DuckLakeIngestionHandler implements IngestionHandler {
     }
 
     @Override
+    public WatermarkSpec getWatermarkSpec(String queueId) {
+        QueueIdToTableMapping mapping = queueIdsToTableMappings.get(queueId);
+        if (mapping == null) mapping = queueIdsToTableMappings.get(extractSuffix(queueId));
+        return mapping == null ? null : WatermarkSpec.fromParameters(queueId, mapping.additionalParameters());
+    }
+
+    @Override
     public PostIngestionTask createPostIngestionTask(IngestionResult result) {
         QueueIdToTableMapping mapping = queueIdsToTableMappings.get(result.queueName());
         if (mapping == null) mapping = queueIdsToTableMappings.get(extractSuffix(result.queueName()));
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakePostIngestionTask.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakePostIngestionTask.java
@@ -7,31 +7,23 @@ import org.slf4j.LoggerFactory;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.ArrayList;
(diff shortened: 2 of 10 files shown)


gtaneja-sentrinox:
Revised in 2f71326e after an adversarial review pass on the first version. The substantive change: watermark rows are now **computed at write time** from the pre-COPY relation and carried through `IngestionResult`, so the registration transaction no longer re-reads the written Parquet. That removes the failure modes the review flagged — schema-strict `read_parquet` binding, a second S3 read inside the transaction, hive-partition column typing, and NULL rows for empty batches — all of which could previously roll back a batch's registration after its inputs were already deleted. Config validation also moved to load time (startup failure with the offending key named, instead of per-batch orphans), identifiers are quoted via the existing `HeaderUtils.quoteIdentifier`, and `ducklake_add_data_files` arguments are now escaped. Full commons suite under JDK 21: 383/383 green.

PR #381, opened 2026-08-19
ci: add release workflow for tags

tanejagagan:
Adds `.github/workflows/release.yml`, automating exactly what the 0.2.14 release did by
hand. Triggered by pushing a `v*` tag (plus `workflow_dispatch` for a re-run).

Best reviewed alongside #380, which adds the PR/push CI. The two touch different files and
do not conflict.

## Three jobs

**`verify`** — resolves the version via `help:evaluate` and refuses to continue if it is a
`-SNAPSHOT`, or if the tag disagrees with the pom (`v0.2.14` must mean pom `0.2.14`). This
guard exists because the release flow alternates between release and snapshot versions on
`main`, so a tag pushed at the wrong commit would otherwise republish `:latest` from a
development build. Then runs the full `clean verify`.

**`publish-images`** — logs in to Docker Hub, installs the reactor, and runs six Jib builds,
then stitches the multi-arch manifests with `docker buildx imagetools create` and verifies
both architectures landed.

**`github-release`** — `gh release create --generate-notes --latest`. Tag pushes only.

## Details worth reviewing

- **The otel image name is overridden deliberately.** `dazzleduck-sql-otel-collector/pom.xml`
  defaults `jib.to.image` to `dazzleduck/dazzleduck-sql-otel-collector`, which is *not* the
  published repo (`dazzleduck/dazzleduck-otel-collector`, no `-sql-`). The workflow passes
  the correct name explicitly. Runtime and compactor need no override — their poms already
  default to the right `<repo>:${project.version}-${jib.architecture}`. Worth fixing the
  otel pom separately so the default is not a trap.
- **Jib cross-builds**, so one amd64 runner produces both architectures — no matrix, no QEMU.
- **The two arches run sequentially**, because the compactor shades a fat jar and concurrent
  builds of that module corrupt `target/`.
- **`latest` always moves** to the released version, matching the current manual behaviour.
  If you ever patch an older line, that would need a guard.

## Verification

I could not execute the workflow (Actions has never run on this repo — see #380), so I
validated what is checkable locally:

- YAML parses; every `run` block passes `bash -n`.
- `./mvnw -q help:evaluate -Dexpression=project.version -DforceStdout` returns exactly
  `0.2.14`, clean on stdout.
- The manifest assertion was tested against the images actually published for 0.2.14:
  it counts 2 for the real multi-arch tags and 0 for a single-arch `0.2.14-amd64` tag,
  so the guard both passes and fails correctly.
- The Jib invocations mirror the commands that produced the live 0.2.14 images.

## Required before this can run

Repository secrets `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`. Without them the Jib step
fails on the private `dazzleduck/base-jre` base image.

## Not included

Maven Central deployment. The `central-publishing-maven-plugin` and the
`release-sign-artifacts` GPG profile are already configured in the root pom, but publishing
to Central is irreversible, so wiring it into an automatic tag trigger is a decision for you
rather than something to slip into this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,173 @@
+name: Release
+
+on:
+  push:
+    tags: ['v*']
+  workflow_dispatch:
+
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+env:
+  JIB_ARGS: -B -ntp -DskipTests
+
+jobs:
+  # Reads the version from the pom and refuses to go further if it is a
+  # SNAPSHOT, or if it disagrees with the tag being released. Without this a
+  # mistagged commit could publish :latest from a development version.
+  verify:
+    name: Verify and resolve version
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    outputs:
+      version: ${{ steps.resolve.outputs.version }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+          cache: 'maven'
+
+      - name: Resolve and validate version
+        id: resolve
+        run: |
+          set -euo pipefail
+          version=$(./mvnw -B -ntp -q help:evaluate -Dexpression=project.version -DforceStdout)
+          echo "Resolved project version: $version"
+
+          case "$version" in
+            *-SNAPSHOT)
+              echo "::error::Refusing to release a SNAPSHOT version ($version)."
+              exit 1
+              ;;
+          esac
+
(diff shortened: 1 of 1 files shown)


tanejagagan:
Rebased onto current `main` (which now carries #380's `ci.yml`) and dropped the
`-Djib.to.image` override for the OTel collector, so all three image names are owned by
their poms.

**Merge order now matters:** this depends on #382, which points the otel pom at
`dazzleduck/dazzleduck-otel-collector`. Without the override, merging this workflow before
#382 would publish the collector to `dazzleduck/dazzleduck-sql-otel-collector` — the wrong,
stale repo. Please merge #382 first.

dazzleduck-web (approved):

PR #389, opened 2026-08-21
feat(ingestion): add MAX timestamp and row count to the watermark row

tanejagagan:
The ingestion watermark carried only `MIN(timestamp)` per group. It now carries **MIN, MAX
and `COUNT(*)`**, computed in the same write-time aggregation and inserted in the same
transaction as the file registration — so the added values inherit the existing atomicity
guarantee for free.

```sql
SELECT "county", MIN("ts") AS "min_ts", MAX("ts") AS "max_ts", COUNT(*) AS "row_count"
FROM (...) GROUP BY "county" HAVING MIN("ts") IS NOT NULL
```

## Configuration

All four keys are required. The source column and the three destinations are now named
separately:

| Key | Role |
|---|---|
| `watermark_timestamp_column` | **source** column in the ingested data; MIN and MAX both read it |
| `watermark_min_timestamp_column` | watermark-table column the MIN lands in |
| `watermark_max_timestamp_column` | watermark-table column the MAX lands in |
| `watermark_row_count_column` | watermark-table column `COUNT(*)` lands in |

Previously `watermark_timestamp_column` did double duty: it named the source column *and*
implicitly the destination for the MIN. That was invisible while MIN was the only aggregate
and became inconsistent once MAX and COUNT had explicit destinations. Renaming it to
`watermark_min_timestamp_column` would have been wrong — MAX reads the same source — so the
source keeps its name and the MIN destination becomes its own key.

Required rather than optional because the feature is not yet in production, so there is no
existing watermark table whose INSERT would break. Existing watermark tables need two new
columns and configs need three new keys; the unknown-key typo guard means a stale config
fails at startup rather than writing to the wrong column.

## Verification

`./mvnw -B -ntp clean verify` on JDK 21: **915 tests, 0 failures, 0 errors**, BUILD SUCCESS.

New and updated coverage in `WatermarkSpecTest` and `DuckLakeWatermarkPostIngestionTest`:

- aggregation always emits MIN, MAX and COUNT, aliased to their configured destinations
- INSERT column order matches aggregate order
- blank or missing MIN/MAX/row-count keys are rejected, at both the record and parse level
- end-to-end: the post-ingestion transaction writes `min|max|count` per group, aggregated
  across **both** batch files rather than per file (`king|wa|01:00|03:00|2`)

Tests deliberately use a MIN destination (`min_ts`) different from the source (`ts`), so a
regression back to the implicit naming would fail.

## Note on an existing behaviour, unchanged

`HAVING MIN(...) IS NOT NULL` still governs which rows are emitted, so a group whose
timestamps are all NULL produces no row at all — meaning its row count is not recorded. That
mattered less when MIN was the only output; now that rows are being counted it is a real
gap. I documented it in the class javadoc rather than changing behaviour in this PR.

### DIFF
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/WatermarkSpec.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/WatermarkSpec.java
@@ -29,8 +29,8 @@ import java.util.stream.Collectors;
  *       {@code COUNT(*)} for the group.</li>
  * </ul>
  *
- * <p>Note that {@code HAVING MIN(..) IS NOT NULL} still governs which rows are emitted, so a
- * group whose timestamps are all NULL produces no row at all — its row count is not recorded.
+ * <p>A group whose timestamps are all NULL still produces a row: NULL min and max, with the real
+ * row count. Only a genuinely empty batch is skipped.
  *
  * <p>Watermark rows are computed at WRITE time by {@link ParquetIngestionQueue} — an aggregation
  * over the same pre-COPY relation the output Parquet is written from (local data, transformation
@@ -39,7 +39,7 @@ import java.util.stream.Collectors;
  * the SAME transaction as the {@code ducklake_add_data_files} registration. The post-ingestion
  * step therefore never re-reads the written files: no second download, no dependence on the
  * written files' schema, and partition columns (present in the relation, hive-only in the files)
- * group correctly. Rows whose MIN would be NULL (empty batch / all-NULL timestamps) are excluded.
+ * group correctly. Only empty batches are excluded.
  *
  * <p>Validation runs at config-load time via {@link QueueIdToTableMapping}, so a malformed spec
  * fails startup rather than orphaning batches at flush time. Values are rejected when blank, and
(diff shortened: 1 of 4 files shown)


tanejagagan:
Pushed a follow-up commit that fixes the gap flagged in the PR description rather than leaving it documented.

**`HAVING MIN(ts) IS NOT NULL` → `HAVING COUNT(*) > 0`**

The old predicate dropped any group whose timestamps were all NULL. That was harmless while MIN was the only output, but with `COUNT(*)` in the row it silently discarded the count for those batches — rows were ingested and never accounted for.

`COUNT(*) > 0` keeps such a group, writing NULL min/max with its true row count, and still suppresses the single all-NULL row a zero-row relation produces in global (ungrouped) mode. So only genuinely empty batches are skipped.

Coverage added:

- `keepsAllNullTimestampGroupSoItsRowCountIsRecorded` — the all-NULL group is now emitted as `[null, null, "1"]` rather than dropped
- `globalModeSkipsOnlyGenuinelyEmptyInput` — empty still yields nothing; all-NULL yields one row
- End-to-end through the post-ingestion transaction: the row lands in the DuckLake watermark table as `king|wa|NULL|NULL|2`, which also proves NULL literals survive the `INSERT ... VALUES` rendering

`./mvnw -B -ntp clean verify` on JDK 21: **916 tests, 0 failures, 0 errors**, BUILD SUCCESS.

PR #394, opened 2026-08-27
feat(otel-collector): cluster scoping via the x-dd-cluster JWT claim

piyushmahoree:
## Summary

Adds **cluster scoping** to the OTel collector: an audience-style check binding JWTs to one cluster, mirroring how the `x-dd-ingestion-queue` claim is checked today.

When `otel_collector.cluster` is configured, every gRPC call must carry a matching `x-dd-cluster` JWT claim; a missing or different claim is rejected with `PERMISSION_DENIED` (distinct from the queue check's `INVALID_ARGUMENT`, since this is an authorization failure on a well-formed request). When the config is unset — the default — the claim is ignored entirely, so existing deployments and gradual rollouts are unaffected.

**Why:** in multi-cluster deployments that share a `secret_key` or a central login service, a token minted for cluster A's collector is otherwise perfectly valid on cluster B's collector. Nothing in the token binds it to a deployment. This closes that replay gap.

| Collector `cluster` config | JWT `x-dd-cluster` claim | Result |
|---|---|---|
| set | matching | allowed (queue check still applies) |
| set | different or absent | rejected — `PERMISSION_DENIED` |
| unset (default) | anything | check skipped — behavior unchanged |

## Changes

- `Headers.CLAIM_CLUSTER` constant (`x-dd-cluster`) in `dazzleduck-sql-common`, next to the queue claim
- `cluster` config key on `CollectorConfig` / `CollectorProperties`, documented in `reference.conf` and the module README (with the rollout matrix)
- Enforcement lives in exactly one place — `JwtServerInterceptor.startAuthorizedCall` — covering both the Bearer path and the Basic path (which mints/receives a token and runs the same authorization on it). A locally issued token whose claims cannot be parsed is now terminal (`UNAUTHENTICATED`) instead of silently proceeding without claims, removing a pre-existing swallowed-exception fall-through.
- A `/simplify` review pass was applied before committing: the `closeCall` helper is now used at all seven rejection sites, the rejection message is unified and names the claim, the startup log uses the `Headers` constant, and the tests reuse the sibling suites' token builder (`tokenWithQueueClaim` gained a cluster overload) and `basicAuthMetadata` helper, with a shared `ServerFixture` base for the two nested classes.

## Design notes

- Claim name follows the project's `x-dd-` convention; single-value config now, shaped so an allowlist (`clusters = [...]`) can drop in later for shared multi-cluster collectors.
- Enforcement is deliberately otel-collector-only for now; the same replay gap exists on HTTP `/v1/ingest` and Flight bulk ingest and can adopt the same claim in a follow-up.
- Locally minted (Basic auth) tokens carry no claims, so under enforcement they are rejected — consistent with the strict contract. Stamping the collector's own cluster into locally issued tokens is a possible follow-up if Basic auth should stay usable under enforcement.

## Test plan

`OtelCollectorClusterClaimTest` covers the full matrix over a real gRPC channel: matching / wrong / missing claim, config-unset ignore (both with and without the claim), cluster check ordering before the queue check (asserted via the claim-naming description), and the Basic-auth path.

Verified locally on JDK 21: all 7 new tests pass; the module suite is green except the pre-existing Windows-only `OtelCollectorConfigDuckLakeTest` failure (raw Windows temp path interpolated into HOCON), which exists on `main` and is unrelated to this change.

### DIFF
--- a/dazzleduck-sql-common/src/main/java/io/dazzleduck/sql/common/Headers.java
+++ b/dazzleduck-sql-common/src/main/java/io/dazzleduck/sql/common/Headers.java
@@ -64,6 +64,10 @@ public class Headers {
     // JWT claim carrying the target ingestion queue ID; same wire value as HEADER_INGESTION_QUEUE.
     public static final String CLAIM_INGESTION_QUEUE = "x-dd-ingestion-queue";
 
+    // JWT claim binding a token to one cluster (audience-style check). A collector configured
+    // with a cluster name rejects tokens whose claim is absent or different.
+    public static final String CLAIM_CLUSTER = "x-dd-cluster";
+
     public static final Set<String> SUPPORTED_HEADERS = Set.of(HEADER_FETCH_SIZE, HEADER_DATABASE, HEADER_SCHEMA, HEADER_SPLIT_SIZE,
             HEADER_DATA_PARTITION, HEADER_DATA_FORMAT, HEADER_PRODUCER_ID, HEADER_PRODUCER_BATCH_ID, HEADER_SORT_ORDER,
             HEADER_APP_DATA_TRANSFORMATION, HEADER_PATH, HEADER_TABLE, HEADER_FUNCTION, HEADER_FILTER, HEADER_ACCESS,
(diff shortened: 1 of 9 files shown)


piyushmahoree:
Withdrawing for now — the design is being reworked together with a follow-up (claims column on ingested rows); will reopen or supersede once that discussion settles.

PR #395, opened 2026-08-31
feat(otel-collector): per-queue claims column from JWT claims (extract_claims)

piyushmahoree:
## Summary

Adds an opt-in **claims column** to the OTel collector: a queue mapping can set `extract_claims = true`, and every row ingested into that queue then carries a `claims MAP(VARCHAR, VARCHAR)` column — alongside `attributes` and `resource_attributes` — populated from the caller's **verified** JWT.

- Registered token claims (`exp`, `iat`, `nbf`, `iss`, `aud`, `jti`) are excluded; `sub` and all custom claims are kept, values stringified.
- The map is stamped **per request at batch-write time** (in the signal services, from claims the auth interceptor already parsed), so it stays correct even when the ingestion queue combines batches from different tokens into one Parquet write — the reason this cannot be done with a per-queue `transformation`, which runs once per bucket.
- Default **off**; zero behavior or schema change for queues that don't opt in.

**Why:** it makes authenticated per-request data available to transformations and partitioning. One shared view can derive a first-class column from any claim:

```sql
CREATE VIEW my_catalog.main.logs_transform AS
SELECT *, claims['org_id'] AS org_id FROM my_catalog.main.raw_logs;

ALTER TABLE my_catalog.main.logs SET PARTITIONED BY (org_id);
```

## Changes

**commons**
- `extract_claims` key on `ingestion_queue_table_mapping` entries; `QueueIdToTableMapping` gains the component plus a `withExtractClaims` wither (existing constructors unchanged, default `false`)
- `IngestionHandler.extractClaims(queueId)` default method and the `CLAIMS_COLUMN` constant; `DuckLakeIngestionHandler` lookup (with a new `mappingFor` helper deduplicating the exact-then-suffix idiom at three pre-existing sites)
- On queue-state build, a **warning when the DuckLake target table lacks the `claims` column** (via `DuckLakeTableManager.currentColumnNames`) — registration tolerates extra file columns, so the data would otherwise be dropped silently; the warning names the exact `ALTER TABLE` to run

**otel-collector**
- `JwtServerInterceptor` publishes the filtered claims in a gRPC context key next to the existing queue-claim key
- The claims schema variant is derived and memoized from each base signal schema, with the column **appended last** so the per-signal batch writers are untouched; `OtelServiceBase` picks the variant per request
- `ClaimsColumnWriter` pre-encodes the constant map to Arrow `Text` once (not once per row) and fills the column after the batch writer runs; a shared `writeEntry` helper now owns the map-writer cast idiom at all five sites in the module
- `reference.conf` + README document the flag, the privacy caveat (enabling it persists token claims into the data), the DDL, and the restart rule

**Limitations (documented):** flag changes require a restart (a bucket must not mix batches with and without the column); the dynamic SQLite registry does not carry the flag — same limitation watermarks have.

## Test plan

- `ExtractClaimsMappingTest` (commons): record defaults, provider config parsing, handler lookup against a real DuckLake catalog (also runs the missing-column warning path)
- `JwtClaimsFilterTest`: registered claims dropped, custom claims kept, non-string values stringified
- `OtelCollectorClaimsColumnTest`: end-to-end over real gRPC for all three signals — Parquet contents queried back via SQL asserting the claim values are present and `exp` is absent; plus the flag-off case asserting the written schema has **no** `claims` column
- Verified locally on JDK 21: all new tests pass; module suites green except the pre-existing Windows-only `OtelCollectorConfigDuckLakeTest` failure (raw Windows temp path in HOCON), which exists on `main` and is unrelated

Independent of #394 (the cluster-scoping PR): this branch is based directly on `main`, deliberately avoids the `x-dd-*` claim namespace in examples/tests, and the two merge cleanly in either order.

### DIFF
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
@@ -356,7 +356,7 @@ public class DuckLakeIngestionHandler implements IngestionHandler {
                     existing.partitionColumns(), existing.partitionProjections(),
                     existing.schemaChangeId(), clock.instant());
         }
-        return buildState(mapping, currentSchemaChangeId, clock.instant());
+        return buildState(mapping, currentSchemaChangeId, clock.instant(), existing == null);
     }
 
     /**
@@ -364,10 +364,13 @@ public class DuckLakeIngestionHandler implements IngestionHandler {
      * Accepts a pre-fetched {@code schemaChangeId} to avoid a redundant round-trip when
      * called from {@link #getOrRefreshState}.
      */
-    private static QueueState buildState(QueueIdToTableMapping mapping, long schemaChangeId, Instant refreshedAt) {
+    private static QueueState buildState(QueueIdToTableMapping mapping, long schemaChangeId, Instant refreshedAt,
+                                         boolean firstBuild) {
         String path           = fetchPath(mapping.catalog(), mapping.schema(), mapping.table());
         String transformation = resolveTransformation(mapping);
-        warnIfClaimsColumnMissing(mapping);
+        if (firstBuild) {
+            warnIfClaimsColumnMissing(mapping, transformation);
+        }
         List<ResolvedPartition> partitions = fetchPartitions(mapping.catalog(), mapping.schema(), mapping.table());
         String[] tokens      = partitions.stream().map(ResolvedPartition::token).toArray(String[]::new);
         String[] projections = partitions.stream().map(ResolvedPartition::projection)
@@ -378,7 +381,7 @@ public class DuckLakeIngestionHandler implements IngestionHandler {
     /** Convenience overload that fetches schema change ID itself (used at construction time). */
     private static QueueState buildState(QueueIdToTableMapping mapping, Instant refreshedAt) {
         long schemaChangeId = fetchSchemaChangeId(mapping.catalog(), mapping.schema(), mapping.table());
-        return buildState(mapping, schemaChangeId, refreshedAt);
+        return buildState(mapping, schemaChangeId, refreshedAt, true);
     }
 
     private static String resolveTransformation(QueueIdToTableMapping mapping) {
@@ -428,11 +431,14 @@ public class DuckLakeIngestionHandler implements IngestionHandler {
     /**
      * DuckLake registration tolerates extra file columns, so with {@code extract_claims} on
      * a target table missing the claims column silently drops the data — warn loudly instead.
+     * A transformation controls the output shape (it may consume claims without persisting
+     * them), so only raw pass-through mappings are checked, and only on the queue's first
+     * state build.
      */
-    private static void warnIfClaimsColumnMissing(QueueIdToTableMapping mapping) {
-        if (!mapping.extractClaims()) return;
+    private static void warnIfClaimsColumnMissing(QueueIdToTableMapping mapping, String transformation) {
+        if (!mapping.extractClaims() || transformation != null) return;
         try (var conn = ConnectionPool.getConnection()) {
-            if (!DuckLakeTableManager.currentColumnNames(conn, mapping).contains(CLAIMS_COLUMN)) {
+            if (!DuckLakeTableManager.hasColumn(conn, mapping, CLAIMS_COLUMN)) {
                 logger.warn("Queue '{}' has extract_claims enabled but table {}.{}.{} has no '{}' column — "
                                 + "claims will be silently dropped at registration. "
                                 + "Run: ALTER TABLE {}.{}.{} ADD COLUMN {} MAP(VARCHAR, VARCHAR)",
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
@@ -180,15 +180,25 @@ public class DuckLakeIngestionHandler implements IngestionHandler {
 
     @Override
     public WatermarkSpec getWatermarkSpec(String queueId) {
-        QueueIdToTableMapping mapping = queueIdsToTableMappings.get(queueId);
-        if (mapping == null) mapping = queueIdsToTableMappings.get(extractSuffix(queueId));
+        QueueIdToTableMapping mapping = mappingFor(queueId);
         return mapping == null ? null : WatermarkSpec.fromParameters(queueId, mapping.additionalParameters());
     }
 
     @Override
+    public boolean extractClaims(String queueId) {
+        QueueIdToTableMapping mapping = mappingFor(queueId);
+        return mapping != null && mapping.extractClaims();
+    }
+
+    /** Mapping for {@code queueId}, resolved exact-first with the path-suffix fallback. */
+    private QueueIdToTableMapping mappingFor(String queueId) {
+        String key = resolveStateKey(queueId);
+        return key == null ? null : queueIdsToTableMappings.get(key);
+    }
+
+    @Override
     public PostIngestionTask createPostIngestionTask(IngestionResult result) {
-        QueueIdToTableMapping mapping = queueIdsToTableMappings.get(result.queueName());
-        if (mapping == null) mapping = queueIdsToTableMappings.get(extractSuffix(result.queueName()));
+        QueueIdToTableMapping mapping = mappingFor(result.queueName());
         if (mapping == null) {
             // No DuckLake mapping for this queue — write-only mode, no catalog registration.
             logger.atDebug().log("No DuckLake mapping for queue '{}', skipping catalog registration", result.queueName());
@@ -335,8 +345,7 @@ public class DuckLakeIngestionHandler implements IngestionHandler {
      * partition columns.
      */
     private QueueState computeRefreshedState(String key, QueueState existing) {
-        QueueIdToTableMapping mapping = queueIdsToTableMappings.get(key);
-        if (mapping == null) mapping = queueIdsToTableMappings.get(extractSuffix(key));
+        QueueIdToTableMapping mapping = mappingFor(key);
         if (mapping == null) return existing; // unknown queue — leave unchanged
 
         long currentSchemaChangeId = fetchSchemaChangeId(mapping.catalog(), mapping.schema(), mapping.table());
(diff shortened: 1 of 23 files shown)


dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
```diff
@@ -415,6 +425,26 @@ static long fetchSchemaChangeId(String catalogName, String schema, String table)
         }
     }
 
+    /**
+     * DuckLake registration tolerates extra file columns, so with {@code extract_claims} on
+     * a target table missing the claims column silently drops the data — warn loudly instead.
+     */
+    private static void warnIfClaimsColumnMissing(QueueIdToTableMapping mapping) {
```
tanejagagan: **The warning misfires on the exact setup the README recommends.**

This checks the *target table* for a `claims` column but ignores the queue's `transformation` / `view`. The README's own privacy guidance steers operators toward a view that consumes claims without persisting them:

```sql
SELECT * EXCLUDE (claims), claims['org_id'] AS org_id FROM __this
```

In that entirely correct configuration the target table legitimately has no `claims` column, so every state rebuild logs a WARN telling the operator to run `ALTER TABLE ... ADD COLUMN claims MAP(VARCHAR, VARCHAR)` — and if they follow that advice, the transformation output no longer matches the table.

Suggest skipping the check when a transformation or view is configured, or running it against the derived transformation output the way `DuckLakeTableManager.deriveSchema` already does.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
```diff
@@ -415,6 +425,26 @@ static long fetchSchemaChangeId(String catalogName, String schema, String table)
...
+        if (!mapping.extractClaims()) return;
+        try (var conn = ConnectionPool.getConnection()) {
+            if (!DuckLakeTableManager.currentColumnNames(conn, mapping).contains(CLAIMS_COLUMN)) {
```
tanejagagan: **Case-sensitive column comparison, inconsistent with the table manager.**

`currentColumnNames(conn, mapping).contains(CLAIMS_COLUMN)` compares case-sensitively, while `DuckLakeTableManager.alterTable` lowercases the same names first (`DuckLakeTableManager::lower`). DuckDB preserves the case of quoted identifiers, so a table whose column was created as `"Claims"` or `"CLAIMS"` reports as missing and fires the warning even though registration will match it fine.

A case-insensitive comparison here keeps the two paths consistent.

dazzleduck-sql-otel-collector/src/main/java/io/dazzleduck/sql/otel/collector/ClaimsColumnWriter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private ClaimsColumnWriter() {}
+
+    static void write(VectorSchemaRoot root, Map<String, String> claims) {
+        int rowCount = root.getRowCount();
+        MapVector claimsVec = (MapVector) root.getVector(IngestionHandler.CLAIMS_COLUMN);
+        UnionMapWriter writer = claimsVec.getWriter();
+
+        // Encode each key/value to UTF-8 once, not once per row.
+        Text[] keys = claims.keySet().stream().map(Text::new).toArray(Text[]::new);
```
tanejagagan: **Keys and values are zipped across two independent map traversals.**

`claims.keySet()` and `claims.values()` are materialized in separate passes and then paired by index. `Map` does not formally guarantee that its `keySet()` and `values()` views iterate in the same order for arbitrary implementations — it holds for the concrete types used today (`LinkedHashMap` from `filterClaims`, `Map.of()`), so this is not a live bug.

But if a future caller ever passes a map whose views diverge, every row silently gets mis-paired keys and values, with no error anywhere to catch it. A single `entrySet()` pass building both arrays removes the coupling at zero cost:

```java
Text[] keys = new Text[claims.size()];
Text[] values = new Text[claims.size()];
int n = 0;
for (Map.Entry<String, String> e : claims.entrySet()) {
    keys[n] = new Text(e.getKey());
    values[n] = new Text(e.getValue());
    n++;
}
```

tanejagagan (commented):
Code review of the claims-column work. Built `dazzleduck-sql-common`, `dazzleduck-sql-commons` and `dazzleduck-sql-otel-collector` on JDK 21 and ran the otel-collector suite (56 tests) plus the commons ingestion suites (128 tests) — all green.

The riskiest part, `ClaimsColumnWriter` writing into a `MapVector` *after* the batch writer has already called `root.setRowCount(n)`, checks out: `ListVector.setValueCount` in Arrow 19 does not touch `lastSet` and recomputes the child count from `offset[lastSet+1]`, and a throwaway test over 5,000 rows with realloc-forcing values, an empty claims map and a zero-row batch read back correctly through `UnionMapReader`. Also confirmed `ducklake_add_data_files` runs with `ignore_extra_columns => true` (so the new warning's premise is right), and that the mixed-schemas-in-one-bucket hazard can't be triggered, since the dynamic path never flips `extractClaims` mid-bucket.

Four low-severity findings inline. Nothing blocking.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionTaskFactoryProvider.java
```diff
@@ -42,9 +42,12 @@ protected Map<String, QueueIdToTableMapping> loadMappings() {
             }
             String view       = c.hasPath(ConfigConstants.VIEW_KEY)        ? c.getString(ConfigConstants.VIEW_KEY)        : null;
             String inputTable = c.hasPath(ConfigConstants.INPUT_TABLE_KEY) ? c.getString(ConfigConstants.INPUT_TABLE_KEY) : null;
+            boolean extractClaims = c.hasPath(ConfigConstants.EXTRACT_CLAIMS_KEY)
```
tanejagagan, line 45: **`extract_claims` is silently ignored under the dynamic provider.**

This is the only place `extract_claims` is read. `DynamicDuckLakeIngestionTaskFactoryProvider.loadMappings()` reads nothing but `db_path`, and `DynamicQueueRepository.loadAll` never sets the flag — so an operator running the dynamic provider who sets `extract_claims = true` under `ingestion_queue_table_mapping` gets no claims column and no diagnostic at all.

The README documents this as a known limitation, but a startup WARN when `extract_claims` appears in config under the dynamic provider would turn a silent misconfiguration into a visible one.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
piyushmahoree: Fixed in 464b0f6e. The warning now takes the transformation already resolved by buildState and returns early when it is non-null, so a mapping that consumes claims without persisting them is never warned about — and the guard cannot drift from transformation resolution. Went with the skip rather than validating derived output: deriveSchema needs inputSchema, which only registry-sourced mappings carry, and extract_claims is HOCON-only today, so the deeper check is unreachable on this path. Also moved the check to the queue's first state build only — it was re-running the information_schema scan on every catalog-wide schema_version bump to repeat an identical warning.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
piyushmahoree: Fixed in 464b0f6e — the comparison now lives in DuckLakeTableManager.hasColumn, built on the manager's existing lower() folding, so both paths share one identifier-matching rule.

dazzleduck-sql-otel-collector/src/main/java/io/dazzleduck/sql/otel/collector/ClaimsColumnWriter.java
piyushmahoree: Fixed in 464b0f6e with your snippet — single entrySet pass building both arrays.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionTaskFactoryProvider.java
piyushmahoree: Fixed in 464b0f6e. The dynamic provider now warns from validate() (which both production call sites already invoke) when the inline mapping sets extract_claims, with the message widened to say the whole inline block is ignored — the registry is the only queue source.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
tanejagagan: Verified fixed in `464b0f6e`. `warnIfClaimsColumnMissing` now takes the resolved transformation and returns early when it is non-null, and `resolveTransformation` covers the `view` case too (via `hasViewTransformation`) — so the README's `EXCLUDE (claims)` view is correctly exempt, not just an explicit `transformation`. The added `firstBuild` gate also stops the warning repeating on every schema-change rebuild, which addresses the log-noise half of this. Thanks.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionHandler.java
tanejagagan: Verified fixed in `464b0f6e`. The check now goes through `DuckLakeTableManager.hasColumn`, which folds both sides with the class's own `lower()`, so the warning path and `alterTable` share one identifier rule. `currentColumnNames` was narrowed to `private` in the same change and has only in-class callers.

dazzleduck-sql-otel-collector/src/main/java/io/dazzleduck/sql/otel/collector/ClaimsColumnWriter.java
tanejagagan: Verified fixed in `464b0f6e` — single `entrySet()` pass building both arrays. Pairing is now structural rather than dependent on view iteration order.

dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakeIngestionTaskFactoryProvider.java
tanejagagan: Verified fixed in `464b0f6e`. The `validate()` override warns when an inline `ingestion_queue_table_mapping` sets `extract_claims` under the dynamic provider, and both production call sites — `FlightSqlProducerFactory.java:562` and `CollectorConfig.java:165` — call `provider.validate()`, so the warning actually reaches an operator at startup.

PR #399, opened 2026-09-02
feat(ingestion): optional watermark_snapshot_id_column

a-chidam:
Records the DuckLake snapshot a batch committed at, on that batch's own watermark row.

Opt-in and additive — without the key nothing changes, and every existing queue configuration keeps its current behaviour. `insertSql` is untouched.

## Why it can't be part of the INSERT

A transaction cannot learn the snapshot it is about to create. Measured on DuckDB 1.5.2 + ducklake:

```
                          ducklake_current_snapshot   ducklake_last_committed_snapshot
before write                        3                            NULL
inside txn, after write             3                            NULL      <- pending id (4) invisible
after commit                        4                            4
```

`current_snapshot()` reports the snapshot being **read**, so inside the ingest transaction it still names the previous one. The new id exists only once the commit completes.

## What this does instead

`DuckLakePostIngestionTask` follows the ingest transaction with one statement, on the **same connection**:

```sql
UPDATE <watermark table>
   SET <snapshot column> = (SELECT * FROM ducklake_last_committed_snapshot('<catalog>'))
 WHERE <snapshot column> IS NULL
```

`ducklake_last_committed_snapshot` is per-connection, and the subquery is evaluated before the `UPDATE`'s own commit — so the value recorded is the **ingest batch's** snapshot, the one that made both the data files and the watermark row visible, and never another writer's. The end-to-end test asserts exactly that: reading the fact table `AT (VERSION => <stamped id>)` already shows the batch's rows.

## Two consequences, documented at both sites

- The stamp is its own transaction, so a queue that configures the key creates **one extra snapshot per batch**.
- A crash between the two commits leaves the row with a NULL id. The `IS NULL` predicate makes the next batch sweep it up, and readers must tolerate NULL in the meantime. That predicate assumes a **single writer per watermark table** — the existing model, one queue to one watermark table.

## Shape of the key

Optional, unlike the other five: absent means the column is never written. Present-but-blank is rejected, since that's a typo rather than a way to disable it. A 6-argument `WatermarkSpec` constructor keeps existing call sites compiling.

## Tests

`WatermarkSpecTest` — parsing (present / absent / blank), the rendered statement, and that `insertSql` still doesn't mention the column.

`DuckLakeWatermarkPostIngestionTest` — two end-to-end cases through `DuckLakePostIngestionTask`: every row carries the same id and it is the snapshot the data landed in; and with the key unconfigured the column stays NULL.

```
WatermarkSpecTest                  17 tests
DuckLakeWatermarkPostIngestionTest  7 tests
                                   24 passed
```

`DuckLakePostIngestionTaskTest.shouldAddFilesToDuckLakeTableTest` fails on my machine with an Arrow/netty `ExceptionInInitializerError` — it fails identically on a clean checkout of `main`, so it's a local JDK/native-access issue, not this change.

## One open question for maintainers

`doc/INGESTION_ARCHITECTURE.md` references a `PRODUCER_WATERMARK_SPEC.md` ("proposed durable producer watermarks") that isn't in the tree. If that design names this key differently, this should conform to it — renaming later is a breaking config change, since an unknown `watermark_`-prefixed key fails startup.

## Motivation

A downstream consumer resumes incremental aggregation from a DuckLake snapshot cursor. When its checkpoint's snapshot expires it needs a second, non-expiring way to bracket unconsumed batches. Today it derives one from the watermark row's `rowid`, which works but can't be compared against a stored snapshot id; with this key the watermark row carries the id directly and the derivation goes away.

### DIFF
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakePostIngestionTask.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DuckLakePostIngestionTask.java
@@ -21,6 +21,13 @@ import java.util.Map;
  * registration and watermark commit or roll back together. This task never re-reads the written
  * files.
  *
+ * <p>When the spec also configures {@code watermark_snapshot_id_column}, a second statement runs
+ * on the SAME connection once that transaction has committed, recording the snapshot the batch
+ * committed at. It cannot be part of the transaction: a transaction cannot learn the snapshot it
+ * is about to create. See {@link WatermarkSpec} for the mechanism and its two consequences — one
+ * extra snapshot per batch, and a NULL that the next batch sweeps up if the process dies between
+ * the two commits.
+ *
  * <p>Limitation: queues registered through the dynamic SQLite registry
  * ({@link DynamicQueueRepository}) do not carry {@code additional_parameters}, so watermarks are
  * only available for statically configured queue mappings.
@@ -81,8 +88,14 @@ public class DuckLakePostIngestionTask implements PostIngestionTask {
         if (watermarkSpec != null && watermarkRows != null && !watermarkRows.isEmpty()) {
             queries.add(watermarkSpec.insertSql(catalogName, schemaName, watermarkRows));
         }
+        String stampSql = watermarkSpec == null ? null : watermarkSpec.snapshotStampSql(catalogName, schemaName);
         try (Connection conn = ConnectionPool.getConnection()) {
             ConnectionPool.executeBatchInTxn(conn, queries.toArray(String[]::new));
+            if (stampSql != null) {
+                // Same connection, after the commit: ducklake_last_committed_snapshot() is
+                // per-connection and now names the transaction above.
+                ConnectionPool.execute(conn, stampSql);
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


a-chidam:
Context on timing, in case it helps you schedule this.

A downstream consumer ([Sentrinox/analytics-schema](https://github.com/Sentrinox/analytics-schema)) is waiting on a published `dazzleduck/dazzleduck-otel-collector` image that includes this, so it can set `watermark_snapshot_id_column` in its collector config. Its config change is already staged and gated on the tag existing.

I've opened #400 with the `0.2.17` version bump (mechanical, mirrors #396) in case that's useful — it's explicitly marked to merge **after** this PR, since merging it first would cut a release without this change. Entirely happy for you to close it and cut the release however you normally do; the tag and publish are yours regardless.

Two questions whenever you get to it:

1. Would you take this for 0.2.17, or would you rather it wait?
2. `doc/INGESTION_ARCHITECTURE.md` references a `PRODUCER_WATERMARK_SPEC.md` that isn't in the tree. If that proposed design names this key something other than `watermark_snapshot_id_column`, I'd rather conform to it now — renaming after release is a breaking config change, since an unknown `watermark_`-prefixed key fails startup.

No rush on either; the consumer works today without the column (it derives an ordering from the watermark row's `rowid`), this just lets it drop that workaround.

dazzleduck-web:
Fixed with other PR

a-chidam:
Superseded by #401, which is merged — closing.

#401 solves it better. Both of us hit the same wall (DuckLake doesn't expose the pending snapshot id; it's assigned at COMMIT), but where I settled for a second transaction stamping the row afterwards, #401 predicts `max(snapshot_id) + 1` before the transaction opens and verifies against `begin_snapshot` after, repairing if a competing writer takes the id first. That keeps file registration, the watermark row and the snapshot id in **one** transaction and one snapshot — no extra commit, and no window where the column reads NULL. It's also verified on DuckDB 1.5.4.0, the version this project ships; I only had 1.5.2 to hand.

Making the key **required** is the right call too. An optional key means a silently unstamped column, which reads exactly like "no gateway ever fell behind" downstream — the failure mode the watermark exists to remove.

Thanks for the quick turnaround on 0.2.17. The downstream consumer is adopting it now, with `watermark_snapshot_id_column = "snapshot_id"` and the base image moving to 0.2.17 in the same commit, since the key is mandatory in both directions.

PR #410, opened 2026-09-11
feat(compactor): export compaction metrics over OTLP gRPC

piyushmahoree:
> **Stacked on #411.** That PR is the first commit on this branch; GitHub shows both until it merges. **Review only the second commit** (`feat(compactor): export compaction metrics over OTLP gRPC`) — 4 files, +152/-4.

## Why

The compactor collects its meters on a `LoggingMeterRegistry`, so nothing ever leaves the process.

## How

Export goes to an OTLP gRPC endpoint through the `micrometer-1.5` bridge, so `CompactionState` keeps its Micrometer instruments rather than being rewritten against the OpenTelemetry metrics API. OTel core is pinned to the version the bridge targets (1.50.0 / 2.16.0-alpha), following the leaf-pom BOM pattern the otel-collector module already uses. The gRPC framing comes from the OkHttp sender pulled in by `exporter-otlp`, so no grpc-java dependency is needed.

Disabled by default, falling back to logging meters — a disabled exporter must not silently discard every meter, which would be worse than the log lines this service emitted before.

## Two load-bearing details, easy to miss on review

- **`ServicesResourceTransformer` in the shade config.** The exporter resolves its gRPC sender through `ServiceLoader`; without merging `META-INF/services` the fat jar starts fine and then fails on the first export. Verified the shaded jar retains `GrpcSenderProvider` -> `OkHttpGrpcSenderProvider`.
- **`endpoint` is the collector's gRPC port**, not its health port — which defaults to 8081 and collides with the runtime's HTTP port.

The configured token must carry the ingestion-queue claim the collector requires; it rejects exports without one and has no fallback.

## Worth discussing

This puts OTel support in one service's module. The alternative is teaching the shared `dazzleduck-sql-micrometer` module about OTLP so other services get it too — but that module is a single-transport Arrow forwarder rather than a pluggable-destination abstraction, so it would mean a second registry plus a destination selector, not an extension point. Happy to go that way instead if preferred.

Also note `opentelemetry-micrometer-1.5` is an alpha artifact (the OTel instrumentation project versions all of its bridges that way).

## Testing

`./mvnw test -pl dazzleduck-sql-ducklake-compactor` -> 16/16 pass (JDK 21). The exporter is off by default, so the suite exercises the logging fallback path.

## Follow-ups, not in this PR

- `FlightSqlProducerFactory` hardcodes `LoggingMeterRegistry` (the same class of gap), and the otel-collector's own meters default to a `SimpleMeterRegistry` with no production caller, so those metrics still go nowhere.


piyushmahoree:
## Verified end to end against a real collector

Ran the full path locally — compactor -> OTLP gRPC -> `dazzleduck-sql-otel-collector` -> Parquet — rather than relying on the unit tests, which exercise the logging fallback (export is off by default) and so never touch the export path.

Setup: collector on `grpc_port = 4317` with a `compactor_metrics` queue; a signed JWT carrying the `x-dd-ingestion-queue` claim; compactor against a real DuckLake catalog seeded with small files, `export_interval = 5s`.

### Happy path

34 successful minor cycles. All ten series arrived, read back out of the Parquet the collector wrote:

| Metric | OTLP type | Attributes | Value |
| --- | --- | --- | --- |
| `ducklake.compaction.minor` | SUM | `{database}` | 34 |
| `ducklake.compaction.failures` | SUM | `{database, type=minor\|major\|housekeeping}` | 0 / 0 / 0 |
| `ducklake.compaction.last_success_age` | GAUGE | `{database}` | 4 |
| `ducklake.compaction.duration` (+`.max`), `ducklake.files.compacted\|small\|medium\|total` | — | — | present |

Counters map to SUM and the age to GAUGE, the Micrometer `type` tag survives the bridge as an OTLP attribute, and `service.name` arrives as a resource attribute.

### Negative control

Same setup pointed at a catalog that was never attached, so every cycle throws. This is the behaviour the change is actually for:

```
failures{type=minor}         1 -> 10     climbing
failures{type=major}         0 ->  0     correctly not blamed
failures{type=housekeeping}  0 ->  0
last_success_age             4 -> 49     climbing, never reset
compaction.minor             0 ->  0     a failed cycle is not a success
```

Before the outcome fix (#411, the first commit on this branch), that same scenario would have shown `minor` incrementing and the timestamp refreshing on every failing cycle.

### Two unrelated problems this turned up

Both filed separately; neither blocks this PR, though the second one affects how the fallback behaves in practice:

- #412 — `-Dconfig.file` is silently ignored by the compactor, because `CompactionConfig` uses `ConfigFactory.load("application")`. My first run fell back to the logging registry because of it.
- #413 — the shaded jar has no SLF4J binding, so it logs nothing. That also makes the `LoggingMeterRegistry` fallback inert in the shipped jar: "export disabled" currently means no telemetry at all, rather than meters in the log.

### Not covered

The manual run is not repeatable in CI. I prototyped an in-process OTLP receiver test that asserts the wire contents (auth header, metric names, `type` attribute) with an explicit `forceFlush` and no sleeps, but it needs ~5 test-scoped gRPC/proto dependencies in this module, so I have left it out of this PR rather than expand its dependency footprint. Happy to add it here or as a follow-up if reviewers want it.

PR #414, opened 2026-09-14
feat(otel-collector): honour temp_write_location for the Arrow scratch directory

gtaneja-sentrinox:
## Why

Each OTLP batch is staged as an **uncompressed** Arrow file on local disk before the Parquet `COPY` reads it. The directory holding those files was created with the single-argument form:

```java
this.tempDir = Files.createTempDirectory(tempDirPrefix);   // OtelServiceBase:67
```

No parent, so it always landed under `java.io.tmpdir`, with **no configuration key** to move it. The only lever was `-Djava.io.tmpdir`, which is awkward in the published image (jib sets no `jvmFlags`) and relocates every other temp file along with it.

That default is a poor fit here: the staging files are uncompressed — larger than the Parquet they become — while `/tmp` is frequently a small `tmpfs`.

The flight module has had a setting for exactly this since before now (`temp_write_location`, `reference.conf:24`), used for the same Arrow staging ahead of the same `COPY`. **The collector was the only ingestion path without one.**

## What

The collector now honours the same key, read through the shared `ConfigConstants.TEMP_WRITE_LOCATION_KEY`:

```hocon
otel_collector {
    temp_write_location = ${java.io.tmpdir}
}
```

HOCON resolves `${java.io.tmpdir}` from system properties, so the key is **always present** — always has a value, while existing deployments keep their current behaviour without setting anything.

The default deliberately differs from flight's `/tmp/dazzleduck-writes`: the collector has no warehouse to sit beside, and that path has the same small-`tmpfs` problem this setting exists to solve.

Each signal service still creates its own scratch directory beneath it:

| Service | Directory prefix |
|---------|------------------|
| logs | `otel-logs-arrow-` |
| traces | `otel-traces-arrow-` |
| metrics | `otel-metrics-arrow-` |

## Startup behaviour

A missing directory is **created**, matching `ConfigConstants.getTempWriteDir` on the flight side — an operator who names a directory gets it, in both modules.

What cannot be made usable fails while the collector is starting, with a message naming the key:

```
otel_collector.temp_write_location exists but is not a directory: /var/data/otel-tmp
```

Without this, a bad value surfaces as an `IOException` on the first export RPC — after the collector is live and telemetry is already flowing. The resolved scratch directory is logged at INFO.

## Unchanged

File lifecycle is untouched: each Arrow file is deleted as soon as its batch is written (`ParquetIngestionQueue:142`), and each scratch directory is removed on shutdown. Steady-state usage stays roughly `ingestion.min_bucket_size` per active queue plus in-flight batches.

This setting controls **only** where batches are staged. The output Parquet location is unaffected — still `output_path` per queue in `ingestion_queue_table_mapping`, or the catalog's own `DATA_PATH` under the DuckLake providers. The README now states that distinction explicitly, since the two are easy to confuse.

## Testing

`CollectorTempWriteLocationTest` (7 tests): the `${java.io.tmpdir}` default resolving through HOCON when unset, an explicit override, the scratch directory being created under the configured path with the right prefix, removal on `close()`, a missing directory being created (including nested), a path that is an existing file being rejected, and a blank value being rejected.

`OtelServiceBaseTest` was updated to pass its JUnit `@TempDir` — a side benefit is that it no longer writes into the real `/tmp` during the build.

```
dazzleduck-sql-otel-collector   63 tests  ✅  (was 56)
dazzleduck-sql-commons         485 tests  ✅
dazzleduck-sql-common           13 tests  ✅
```

Branched off `main` at `dcfba3e1`.

## Commits

1. `4e913f56` — make the scratch directory configurable
2. `7cd78700` — name the key `temp_write_location` and create it if absent, matching the flight module


gtaneja-sentrinox:
Auto-closed by a head-branch rename (`feat/otel-collector-temp-path` → `feat/otel-collector-temp-write-location`, to match the config key). Continued in #415 with the same two commits.

PR #415, opened 2026-09-14
feat(otel-collector): honour temp_write_location for the Arrow scratch directory

gtaneja-sentrinox:
> Replaces #414, which GitHub auto-closed when its head branch was renamed.

## Why

Each OTLP batch is staged as an **uncompressed** Arrow file on local disk before the Parquet `COPY` reads it. The directory holding those files was created with the single-argument form:

```java
this.tempDir = Files.createTempDirectory(tempDirPrefix);   // OtelServiceBase
```

No parent, so it always landed under `java.io.tmpdir`, with **no configuration key** to move it. The only lever was `-Djava.io.tmpdir`, which is awkward in the published image (jib sets no `jvmFlags`) and relocates every other temp file along with it.

That default is a poor fit here: the staging files are uncompressed — larger than the Parquet they become — while `/tmp` is frequently a small `tmpfs`.

The flight module has had a setting for exactly this (`temp_write_location`), used for the same Arrow staging ahead of the same `COPY`. **The collector was the only ingestion path without one.**

## What

The collector now honours the same key, read through the shared `ConfigConstants.TEMP_WRITE_LOCATION_KEY`:

```hocon
otel_collector {
    temp_write_location = ${java.io.tmpdir}"/dazzleduck-writes"
}
```

HOCON resolves `${java.io.tmpdir}` from system properties, so the key is **always present** — always has a value, while existing deployments keep their current behaviour without setting anything.

Both modules now declare the **same default**. On Linux `java.io.tmpdir` is `/tmp`, so that is byte-identical to flight's previous hardcoded `"/tmp/dazzleduck-writes"` — no change for containers, which is where it matters — while off Linux it stops assuming a POSIX `/tmp` and follows `-Djava.io.tmpdir`.

Each signal service still creates its own scratch directory beneath it:

| Service | Directory prefix |
|---------|------------------|
| logs | `otel-logs-arrow-` |
| traces | `otel-traces-arrow-` |
| metrics | `otel-metrics-arrow-` |

## Where the work happens

`OtelCollectorServer` owns the whole directory lifecycle. `start()` resolves `temp_write_location` **once, before anything is allocated**, then creates one scratch directory per signal and passes each in already made:

```java
Path tempWriteDir = ConfigConstants.getTempWriteDir(props.getTempWriteLocation());
...
logService = new OtelLogService(
        createScratchDir(tempWriteDir, OtelLogService.SCRATCH_PREFIX), ...);
```

Validating and creating operator config is a startup concern, not a per-service one — previously all three services passed the raw config string down and re-created and re-validated the same parent.

**The service constructors now do no I/O at all.** `createTempDirectory` had been the only reason the base and all three signal-service constructors declared `throws IOException`, and the only reason the allocator had to be constructed in a particular order — a failing creation would otherwise strand a `RootAllocator` that was never assigned to a field and so never closed. Moving the I/O out removes that hazard structurally rather than ordering around it.

`close()` deletes the directories after the services are closed. Each service used to remove its own, which left a gap: if the second of three creations failed, the first directory had no service to clean it up. The server records what it creates, and `start()`'s catch calls `close()`, so a partial startup no longer strands one.

A bad value now fails before the meter registry, health reporter and flush scheduler are built:

```
temp_write_location exists but is not a directory: /var/data/otel-tmp
```

Without this, it surfaces as an `IOException` on the first export RPC — after the collector is live and telemetry is already flowing. The resolved scratch directory is logged at INFO.

## Shared validation (affects `dazzleduck-sql-flight`)

The blank / not-a-directory / not-writable checks live in `ConfigConstants.getTempWriteDir`, behind a new `String` overload, rather than in a collector-private copy. One implementation, so the two modules cannot drift.

**This changes flight's behaviour**: it now fails at startup on an unusable `temp_write_location` instead of failing per write. That seems right — and it is the reason the shared helper was the correct home — but it reaches beyond the collector and deserves an explicit look. Flight's 148 tests pass unchanged.

## Unchanged

File lifecycle is untouched: each Arrow file is deleted as soon as its batch is written (`ParquetIngestionQueue:142`), and each scratch directory is removed on shutdown. Steady-state usage stays roughly `ingestion.min_bucket_size` per active queue plus in-flight batches.

This setting controls **only** where batches are staged. The output Parquet location is unaffected — still `output_path` per queue in `ingestion_queue_table_mapping`, or the catalog's own `DATA_PATH` under the DuckLake providers. The README states that distinction explicitly, since the two are easy to confuse.

## Testing

- `ConfigConstantsTempWriteDirTest` (5 tests, `dazzleduck-sql-common`) — the shared validation, in the module that owns it: creating a missing directory, accepting an existing writable one, and rejecting a blank value, an existing file, and a path whose message must name the key an operator has to fix.
- `CollectorTempWriteLocationTest` (4 tests) — the config value and the server's directory handling: the default resolving through HOCON when unset, an explicit override, the server creating exactly one correctly-prefixed directory per signal under `temp_write_location`, and — the property this refactor is really about — constructing a service against a path that does not exist succeeding and creating nothing.

`OtelServiceBaseTest` was updated to pass its JUnit `@TempDir` — a side benefit is that it no longer writes into the real `/tmp` during the build.

```
dazzleduck-sql-common           18 tests  ✅  (was 13)
dazzleduck-sql-otel-collector   60 tests  ✅  (was 56)
dazzleduck-sql-flight          148 tests  ✅
dazzleduck-sql-commons         485 tests  ✅
```

Branched off `main` at `dcfba3e1`.

## Commits

1. `4e913f56` — make the scratch directory configurable
2. `7cd78700` — name the key `temp_write_location`, create it if absent
3. `1928f032` — give both modules the same default
4. `a8359367` — code review: allocator leak on the fail-fast paths, `CollectorProperties` default, validation moved to the shared helper, orphaned javadoc
5. `d6b0fa83` — resolve once in the server; services take a validated `Path`
6. `537d9d5b` — take filesystem I/O out of the service constructors; server owns create and delete


tanejagagan:
## Code review — `temp_write_location` in the OTel collector

Overall clean and well-tested. Config substitution, directory creation, and cleanup logic are correct; the flight `reference.conf` change to `${java.io.tmpdir}"/dazzleduck-writes"` resolves fine. No high-severity functional bugs. Findings, most concrete first:

**1. RootAllocator leak on the new fail-fast paths** (correctness) — `OtelServiceBase.java:73`
`new RootAllocator()` runs on line 73, then line 74 calls the new temp-dir validation which can throw `IOException` before the allocator is stored in a field — so it is never `close()`'d. This is exactly the blank / not-a-directory / non-writable cases the PR adds (and that the new negative tests exercise), so a previously-unreachable leak becomes routine. Construct/validate the temp dir before allocating, or wrap in a try/catch that closes on failure.

**2. `CollectorProperties` default diverges from the documented default** — `CollectorProperties.java:39`
The field defaults `tempWriteLocation` to bare `java.io.tmpdir`, but the reference.conf / documented default is `java.io.tmpdir/dazzleduck-writes`. Any `CollectorProperties` built directly (several tests do `new CollectorProperties()`) writes scratch dirs straight into `/tmp` instead of `/tmp/dazzleduck-writes`.

**3. Validation added only to the collector, not the shared helper** (altitude) — `OtelServiceBase.java:176`
The fail-fast checks live in the collector's private `resolveTempParent`, while the shared `ConfigConstants.getTempWriteDir` (used by the flight module) still only does exists→createDirectories with no writability/is-directory check. The two paths will drift, and the flight server keeps failing per-write on a bad location. Consider hardening `getTempWriteDir` and delegating to it.

**4. Orphaned Javadoc** (readability) — `OtelServiceBase.java:168`
`resolveTempParent` was inserted between `writeArrowFile`'s Javadoc and its body, so two Javadoc blocks now stack before `resolveTempParent` and `writeArrowFile` is left undocumented.

**5. Misleading "matches getTempWriteDir" claim** — `OtelServiceBase.java:169`
The Javadoc/README assert parity with the flight helper, but the collector actually adds blank/is-directory/is-writable checks the shared helper lacks — so they are not equivalent.

Finding 1 is the one worth fixing before merge.

gtaneja-sentrinox:
All five addressed in `a8359367`. Each was reproduced in the code before fixing.

**1. RootAllocator leak — fixed.** Confirmed: `new RootAllocator()` ran before the validation that can throw, so the allocator was never assigned to a field and never closed. Worse than a latent leak — the negative tests this PR adds exercise exactly those paths, so it would have leaked on every rejection. The scratch directory is now resolved first and the allocator created only once that succeeds.

**2. `CollectorProperties` default — fixed.** It still returned bare `java.io.tmpdir` after the default was aligned; anything constructing `CollectorProperties` directly staged batches loose in the temp root instead of `dazzleduck-writes/`.

**3 + 5. Validation moved to the shared helper.** These were one problem: the checks lived in a collector-private copy, and the javadoc's parity claim was therefore false. Rather than correct the wording, the blank / is-directory / is-writable checks now live in `ConfigConstants.getTempWriteDir` behind a new `String` overload, and the collector delegates. The claim is true because there is one implementation.

Worth flagging as a **behaviour change for the flight server**: it now fails at startup on an unusable `temp_write_location` rather than per write. That is the intent of the finding, and I think it is right — but it is beyond the collector, so it deserves an explicit look. Flight's 148 tests pass unchanged.

**4. Orphaned javadoc — fixed.** Removing the collector-local helper reattached `writeArrowFile`'s javadoc, which the insertion had split off.

New `ConfigConstantsTempWriteDirTest` (5 tests) covers the validation in the module that now owns it, including that the message names the key an operator has to fix.

```
dazzleduck-sql-common           18 tests  ✅  (was 13)
dazzleduck-sql-otel-collector   63 tests  ✅
dazzleduck-sql-flight          148 tests  ✅
dazzleduck-sql-commons         485 tests  ✅
```

gtaneja-sentrinox:
Five of six addressed in `cd1f9908`; the sixth needs a decision rather than a patch.

**Fixed**

- **`trim()` mismatch** — the real one. The check used `trim()` but `Path.of` got the untrimmed string, so `"  /var/data/otel-tmp"` passed validation and became a *relative* path whose first element is `" "`, staging batches under the working directory instead of where the config said. Trimmed once, used once.
- **Flight's wrapper message** — `"Failed to create temp write directory"` overwrote a message that names the key, and two of the three checks aren't creation failures. Passed through.
- **Triple-spelled default** — pinned with a test. It caught a real discrepancy immediately: `CollectorConfig` returns the raw HOCON string while `CollectorProperties` normalises via `Path.of`, so on macOS they differed by the redundant separator `${java.io.tmpdir}` leaves. Same directory, different spelling — the test now compares `Path`s, which is the property that matters.
- **Leaked `DirectoryStream`** in the new server test — closed.
- **Unused `ConfigConstants` / `Comparator` imports** — removed.

**Not fixed — worth a deliberate call**

The predictable-path point on `/tmp/dazzleduck-writes`. You're right that it is weaker than the old `Files.createTempDirectory(prefix)`: the parent is now a fixed name created with umask permissions and a symlink-following existence check, so on a shared host with a sticky world-writable `/tmp` a local user can pre-create it, and the new fail-fast behaviour turns that from degradation into a refusal to start.

I've left it because it is the direct consequence of a deliberate decision in this PR — matching flight's existing key *and* default — and hardening only the collector would re-introduce exactly the divergence the last two commits removed. The per-service scratch directories underneath are still `createTempDirectory`, so they keep `0700`; it is only the shared parent that is predictable, and in a container it is not reachable by another user.

If it should be hardened, the right shape is `PosixFilePermissions` 0700 on create plus `NOFOLLOW_LINKS` on the existence check, applied in `ConfigConstants.getTempWriteDir` so both modules get it — a separate change, since it alters flight's behaviour too.

```
dazzleduck-sql-common           18 tests  ✅
dazzleduck-sql-otel-collector   61 tests  ✅
dazzleduck-sql-flight          148 tests  ✅
dazzleduck-sql-commons         485 tests  ✅
```

PR #416, opened 2026-09-14
feat(auth): carry DuckDB session variables in the JWT (x-dd-variables)

tanejagagan:
## Why

Row-level-security filters today bake the tenant value into the filter string carried in `x-dd-filter` / `x-dd-access` (e.g. `tenant_id = 'acme'`). This adds a new `x-dd-variables` JWT claim so the concrete value travels as **data** in the token, and filters reference it via `getvariable('name')`.

## What

A new `x-dd-variables` claim — a JSON object of string key/values — is applied to the per-request DuckDB connection as `SET VARIABLE`, so both queries and injected RLS filters can read it with `getvariable('name')`.

```
x-dd-variables = {"tenant_id":"acme","region":"us-east"}
```
→ on the connection:
```sql
SET VARIABLE tenant_id = 'acme'
SET VARIABLE region = 'us-east'
```
An `x-dd-access` filter can then read it: `["table","orders","*","tenant_id = getvariable('tenant_id')"]`.

### Trust model
Read **only** from the verified (signed) token — intentionally **not** in `Headers.SUPPORTED_HEADERS`, so a client cannot override it via a request header.

### Format & validation
- Values must be **quoted JSON strings**; a bare number/boolean (`{"n":42}`) is rejected with a hint to quote it (all variables are VARCHAR — cast in filters, e.g. `getvariable('n')::INT`).
- Variable names must match `[A-Za-z_][A-Za-z0-9_]*`; values are rendered as escaped single-quoted literals, so a crafted claim cannot inject SQL.

## How

Applied in `DuckDBFlightSqlProducer.getConnection` — the single chokepoint every per-request connection flows through for **both Flight and HTTP** (HTTP via `SyntheticFlightContext`), planning and execution. No new call sites.

- New `SessionVariables` helper (parse / validate / render) + 11 unit tests.
- `x-dd-variables` added to `jwt_token.claims.generate.headers` (flight + login `reference.conf`).
- Integration tests over **Flight** and **HTTP**: `getvariable()` in a plain query and in an RLS filter.
- Documented in `CLAUDE.md`.

## Testing

`dazzleduck-sql-commons` (SessionVariablesTest 11/11), `dazzleduck-sql-flight` (150), `dazzleduck-sql-http` (107) — all pass on JDK 21. (The `HivePartitionPruningTest` Testcontainers case is skipped locally — needs Docker — and is unrelated to this change.)

https://claude.ai/code/session_011XsgqTSQPrivcvM2UfR22n


gtaneja-sentrinox:
## Review: split planning drops the session variables, so a `getvariable()` RLS filter prunes away every file

The PR says `getConnection` is "the single chokepoint every per-request connection flows through". It isn't — split planning is a second path, and it doesn't carry the variables.

`RestrictedFlightSqlProducer.getFlightInfoStatementSplittable` hands the **already-authorized** tree (RLS filter injected) to `SplitPlanner`, and the partition pruners open their own connections with a bare `ConnectionPool.getConnection()`:

- `HivePartitionPruning.pruneFiles` — `writeConnection`, which runs the pruning query
- `DucklakePartitionPruning.pruneFiles` — the connection that runs `toRun`

Neither ran the request's `SET VARIABLE` statements. Confirmed against DuckDB that `getvariable('never_set') IS NULL` is `true`, so `dt = getvariable('d')` evaluates to NULL and **every file is pruned**.

**Repro:** RESTRICTED mode, `x-dd-access = [["table","orders","*","dt = getvariable('d')"]]`, `orders` partitioned on `dt`. Send the query with `x-dd-split-size` (Flight) or to `/v1/plan` (HTTP) — empty result instead of the tenant's rows. Fails closed, so it is wrong results rather than a leak, but it hits exactly the pattern this PR documents in `CLAUDE.md`. The added tests only cover the non-split path.

### Fix

Thread the request's `SET VARIABLE` statements from the producer down to the connection in each pruner that actually evaluates the filter. 8 files, +172/-15.

- `sessionSetupSqls(CallContext)` extracted on `DuckDBFlightSqlProducer` so `getConnection` and the split path share one source and cannot drift.
- New `List<String> sessionSetupSqls` parameter on `SplitPlanner` / `PartitionPrunerV2`. **Every prior signature is kept as a `List.of()` overload**, so no existing caller or test changed — `SplitPlannerTest`, `HivePartitionPruningTest`, `DuckLakePartitionPruningTest` are untouched.
- Only the connections that evaluate the filter get the variables. The parse-only `readConnection` in the Hive path and the cached metadata lookups in the DuckLake path are deliberately left alone.
- **Delta cannot be fixed this way** — `PartitionPruning` converts the filter to a Delta Kernel `Predicate`, which has no notion of DuckDB session variables and nowhere to apply them. `DeltaLakeSplitPlanner` now detects `getvariable` and throws a clear error instead of handing the Kernel an expression it will mis-evaluate.
- Also folded in: `ConnectionPool.getConnection(String[])` now closes the connection when its setup batch throws. That batch carries caller-influenced SQL on more paths after this change, and without it a rejected `SET VARIABLE` leaked a DuckDB connection per request.

### Test

`SplitPlannerSessionVariableTest` covers both directions. The second test pins the failure mode itself — it passes on the current PR head, which is the proof the bug is real:

```java
@Test
public void withoutTheSessionVariableEveryFileIsPruned() throws SQLException, IOException {
    var statuses = SplitPlanner.getSplitStatus(
            Transformations.parseToTree(QUERY_FILTERED_BY_VARIABLE), 1024 * 1024 * 1024, List.of());
    Assertions.assertEquals(0, statuses.stream().flatMap(List::stream).count());
}
```

With the variables threaded through, the same query returns exactly the two `dt=2025-01-01` files.

Verified on JDK 21 against `6ceaedae`: `dazzleduck-sql-commons` 498 tests, `dazzleduck-sql-flight` 151 tests, 0 failures (skips are the pre-existing Testcontainers ones).

### Not addressed here

This makes the `getvariable()` RLS pattern **work** on the split path; it does not make it **safe**. `x-dd-variables` ships in the default `claims.generate.headers`, and `AdvanceJWTTokenAuthenticator` copies it straight from client connection headers into the signed token, so the value is client-authored — a client can pick its own `tenant_id`. `SessionVariables.validate()` is still an empty TODO. Separate discussion, but worth settling before this lands as the recommended RLS pattern.

Also minor: `dazzleduck-sql-http/reference.conf` now duplicates `jwt_token.claims.generate.headers`, which `dazzleduck-sql-flight/reference.conf` already defines. HOCON lists don't concatenate across `reference.conf` resources — one wins by unspecified classpath order. Identical today, but the next claim added to only one copy will silently vanish.

### Patch

Apply with `git am`:

<details>
<summary><code>0001-fix-flight-apply-session-variables-when-pruning-parti.patch</code></summary>

```diff
From 4ebf72fd93c8053c7ae79d1599729c5b623d8cfd Mon Sep 17 00:00:00 2001
From: Gagan Taneja <gtaneja@sentrinox.ai>
Date: Mon, 14 Sep 2026 12:09:17 -0700
Subject: [PATCH] fix(flight): apply session variables when pruning partitions
 for splits
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

getConnection is not the only connection that evaluates a request's query.
Split planning hands the already-authorized tree — row-level-security filter
injected — to SplitPlanner, and the partition pruners open their own
connections via a bare ConnectionPool.getConnection(). Those connections never
ran the request's SET VARIABLE statements, so a filter of the form
`dt = getvariable('d')` evaluated against NULL and pruned away every file: any
query with x-dd-split-size, or any /v1/plan request, silently returned an empty
result instead of the caller's rows.

Thread the request's SET VARIABLE statements from the producer down through
SplitPlanner and PartitionPrunerV2 to the connection in each pruner that
actually evaluates the filter — writeConnection in the Hive path, the toRun
connection in the DuckLake path. The parse-only and cached-metadata connections
are deliberately left alone. Every prior signature is kept as a List.of()
overload, so no existing caller changes.

The Delta path cannot be fixed the same way: it converts the filter to a Delta
Kernel Predicate, which has no notion of DuckDB session variables and nowhere
to apply them. Detect getvariable() there and fail with a clear message rather
than hand the Kernel an expression it will mis-evaluate.

Also close the connection when ConnectionPool.getConnection(String[]) fails its
setup batch. That batch now carries caller-influenced SQL on more code paths,
and without this a rejected SET VARIABLE leaked a DuckDB connection per request.

SplitPlannerSessionVariableTest covers both directions, including a test that
pins the failure mode itself: the same query with no variables prunes to zero
files.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
---
 .../sql/commons/ConnectionPool.java           | 21 +++++++-
 .../ducklake/DucklakePartitionPruning.java    | 15 +++++-
 .../commons/hive/HivePartitionPruning.java    | 30 +++++++++++-
 .../commons/planner/PartitionPrunerV2.java    | 24 ++++++---
 .../sql/commons/planner/SplitPlanner.java     | 26 +++++++++-
 .../SplitPlannerSessionVariableTest.java      | 49 +++++++++++++++++++
 .../server/DuckDBFlightSqlProducer.java       | 17 ++++++-
 .../server/RestrictedFlightSqlProducer.java   |  5 +-
 8 files changed, 172 insertions(+), 15 deletions(-)
 create mode 100644 dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java

diff --git a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ConnectionPool.java b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ConnectionPool.java
index 7b50a7fc..6b2d920d 100644
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ConnectionPool.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ConnectionPool.java
@@ -426,10 +426,29 @@ public enum ConnectionPool {
      */
     public static DuckDBConnection getConnection(String[] sqls) {
         DuckDBConnection connection = getConnection();
-        executeBatch(connection, sqls);
+        try {
+            executeBatch(connection, sqls);
+        } catch (RuntimeException e) {
+            // The setup batch can fail on caller-influenced SQL (SET VARIABLE from a JWT claim).
+            // Without this the connection we just opened is never closed.
+            try {
+                connection.close();
+            } catch (SQLException suppressed) {
+                e.addSuppressed(suppressed);
+            }
+            throw e;
+        }
         return connection;
     }
 
+    /**
+     * @param sqls Sql which will be executed on connection before connection is returned.
+     * @see #getConnection(String[])
+     */
+    public static DuckDBConnection getConnection(List<String> sqls) {
+        return getConnection(sqls.toArray(new String[0]));
+    }
+
     /**
      *
      * @param reader
diff --git a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ducklake/DucklakePartitionPruning.java b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ducklake/DucklakePartitionPruning.java
index 75d357a4..25d7364a 100644
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ducklake/DucklakePartitionPruning.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ducklake/DucklakePartitionPruning.java
@@ -152,6 +152,19 @@ public class DucklakePartitionPruning {
     public List<FileStatus> pruneFiles(String schema,
                                        String table,
                                        JsonNode tree) throws SQLException {
+        return pruneFiles(schema, table, tree, List.of());
+    }
+
+    /**
+     * @param sessionSetupSqls the request's {@code SET VARIABLE} statements. {@code toRun} carries
+     *                         the caller's filter, so a filter referencing {@code getvariable('x')}
+     *                         evaluates against NULL — pruning away every file — unless the
+     *                         connection that runs it has applied them.
+     */
+    public List<FileStatus> pruneFiles(String schema,
+                                       String table,
+                                       JsonNode tree,
+                                       List<String> sessionSetupSqls) throws SQLException {
         var where = Transformations.identity()
                 .andThen(Transformations::getFirstStatementNode)
                 .andThen(Transformations::getWhereClauseForBaseTable)
@@ -188,7 +201,7 @@ public class DucklakePartitionPruning {
                 toRun = partitionSql(partitionSql, tableId, columnIds);
             }
         }
-        try (var connection = ConnectionPool.getConnection()) {
+        try (var connection = ConnectionPool.getConnection(sessionSetupSqls)) {
             var res = ConnectionPool.collectAll(connection, toRun, DucklakeFileStatus.class);
             var result = new java.util.ArrayList<FileStatus>();
             for (var x : res) {
diff --git a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/hive/HivePartitionPruning.java b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/hive/HivePartitionPruning.java
index 31b15c69..27a09c42 100644
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/hive/HivePartitionPruning.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/hive/HivePartitionPruning.java
@@ -118,6 +118,19 @@ public class HivePartitionPruning extends PartitionPruning {
     public static List<FileStatus> pruneFiles(String basePath,
                                               String filterExpression,
                                               String[][] partitionDataTypes) throws SQLException, IOException {
+        return pruneFiles(basePath, filterExpression, partitionDataTypes, List.of());
+    }
+
+    /**
+     * @param sessionSetupSqls the request's {@code SET VARIABLE} statements. The pruning query
+     *                         carries the caller's filter, so a filter referencing
+     *                         {@code getvariable('x')} evaluates against NULL — pruning away every
+     *                         file — unless the connection that runs it has applied them.
+     */
+    public static List<FileStatus> pruneFiles(String basePath,
+                                              String filterExpression,
+                                              String[][] partitionDataTypes,
+                                              List<String> sessionSetupSqls) throws SQLException, IOException {
         if (partitionDataTypes == null || partitionDataTypes.length == 0) {
             return pruneFilesNoPartition(basePath);
         }
@@ -128,7 +141,9 @@ public class HivePartitionPruning extends PartitionPruning {
             String partitionSql = HivePartitionPruning.getPartitionSql(partitionDataTypes, tempTableName, filterExpression);
             String transformed = doQueryTransformation(readConnection, partitionSql,
                     Arrays.stream(partitionDataTypes).map(ss -> ss[0]).collect(Collectors.toSet()));
-            try (DuckDBConnection writeConnection = ConnectionPool.getConnection();
+            // writeConnection is the one that evaluates `transformed` (the filter), so it — not the
+            // parse-only readConnection — is the connection that needs the session variables.
+            try (DuckDBConnection writeConnection = ConnectionPool.getConnection(sessionSetupSqls);
                  BufferAllocator allocator = new RootAllocator();
                  ArrowReader reader1 = ConnectionPool.getReader(readConnection, allocator, firstSql, 1000);
                  Closeable ignored = ConnectionPool.createTempTableWithMap(writeConnection, allocator, reader1,
@@ -154,7 +169,18 @@ public class HivePartitionPruning extends PartitionPruning {
                                               JsonNode tree,
                                               String[][] partitionDataTypes) throws SQLException, IOException {
 
-        return pruneFiles(basePath, getFilterSql(tree), partitionDataTypes);
+        return pruneFiles(basePath, tree, partitionDataTypes, List.of());
+    }
+
+    /**
+     * @param sessionSetupSqls see {@link #pruneFiles(String, String, String[][], List)}
+     */
+    public static List<FileStatus> pruneFiles(String basePath,
+                                              JsonNode tree,
+                                              String[][] partitionDataTypes,
+                                              List<String> sessionSetupSqls) throws SQLException, IOException {
+
+        return pruneFiles(basePath, getFilterSql(tree), partitionDataTypes, sessionSetupSqls);
     }
     /**
      * Retrieves all files from a specified not partitioned directory path and returns their names and sizes.
diff --git a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/PartitionPrunerV2.java b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/PartitionPrunerV2.java
index ca570a29..d37fa564 100644
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/PartitionPrunerV2.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/PartitionPrunerV2.java
@@ -59,7 +59,8 @@ public interface PartitionPrunerV2 {
      */
     List<FileStatus> pruneFiles(JsonNode tree,
                                 long maxSplitSize,
-                                Map<String, String> properties) throws SQLException, IOException;
+                                Map<String, String> properties,
+                                List<String> sessionSetupSqls) throws SQLException, IOException;
 
 
     static String getPath(JsonNode tree) {
@@ -74,21 +75,31 @@ public interface PartitionPrunerV2 {
 class HiveSplitPlanner implements PartitionPrunerV2 {
 
     @Override
-    public List<FileStatus> pruneFiles(JsonNode tree, long maxSplitSize, Map<String, String> properties) throws SQLException, IOException {
+    public List<FileStatus> pruneFiles(JsonNode tree, long maxSplitSize, Map<String, String> properties,
+                                       List<String> sessionSetupSqls) throws SQLException, IOException {
         var partitionDataTypes  = Transformations.getHivePartition(tree);
         var path  = PartitionPrunerV2.getPath(tree);
         return HivePartitionPruning.pruneFiles(path,
-                tree, partitionDataTypes);
+                tree, partitionDataTypes, sessionSetupSqls);
     }
 }
 
 class DeltaLakeSplitPlanner implements PartitionPrunerV2 {
 
     @Override
-    public List<FileStatus> pruneFiles(JsonNode tree, long maxSplitSize, Map<String, String> properties) throws SQLException, IOException {
+    public List<FileStatus> pruneFiles(JsonNode tree, long maxSplitSize, Map<String, String> properties,
+                                       List<String> sessionSetupSqls) throws SQLException, IOException {
         var statement = Transformations.getFirstStatementNode(tree);
         var path  = PartitionPrunerV2.getPath(tree);
         var filterExpression = Transformations.getWhereClauseForTableFunction(statement);
+        // The Delta path converts the filter to a Delta Kernel Predicate, which has no notion of
+        // DuckDB session variables — there is nowhere to apply sessionSetupSqls. Fail loudly rather
+        // than hand the Kernel an expression it will mis-evaluate.
+        if (filterExpression != null
+                && !Transformations.collectFunction(filterExpression, "getvariable").isEmpty()) {
+            throw new SQLException("getvariable() is not supported in a filter over read_delta; "
+                    + "split planning for Delta cannot evaluate session variables");
+        }
         return PartitionPruning.pruneFiles(path, filterExpression);
     }
 }
@@ -97,7 +108,8 @@ class DucklakeSplitPlanner implements PartitionPrunerV2 {
 
     private Map<String, DucklakePartitionPruning> cache = new ConcurrentHashMap<>();
     @Override
-    public List<FileStatus> pruneFiles( JsonNode tree, long maxSplitSize, Map<String, String> properties) throws SQLException, IOException {
+    public List<FileStatus> pruneFiles( JsonNode tree, long maxSplitSize, Map<String, String> properties,
+                                        List<String> sessionSetupSqls) throws SQLException, IOException {
         var catalogSchemaAndTables =
                 Transformations.getAllTablesOrPathsFromSelect(Transformations.getFirstStatementNode(tree), null, null);
         var first = catalogSchemaAndTables.get(0);
@@ -110,7 +122,7 @@ class DucklakeSplitPlanner implements PartitionPrunerV2 {
                 return oldValue;
             }
         });
-       return pruner.pruneFiles(first.schema(), first.tableOrPath(), tree);
+       return pruner.pruneFiles(first.schema(), first.tableOrPath(), tree, sessionSetupSqls);
 
     }
 }
\ No newline at end of file
diff --git a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/SplitPlanner.java b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/SplitPlanner.java
index 94a50963..c28a7b7d 100644
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/SplitPlanner.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/planner/SplitPlanner.java
@@ -18,6 +18,19 @@ public interface SplitPlanner {
 
     static List<List<FileStatus>> getSplitStatus(JsonNode tree,
                                                  long maxSplitSize) throws SQLException, IOException {
+        return getSplitStatus(tree, maxSplitSize, List.of());
+    }
+
+    /**
+     * @param sessionSetupSqls statements (today the request's {@code SET VARIABLE}s) that every
+     *                         connection evaluating the tree must run first. The tree handed here
+     *                         already carries the injected row-level-security filter, so a filter
+     *                         referencing {@code getvariable('x')} prunes against NULL — i.e.
+     *                         discards every file — on a connection that has not run them.
+     */
+    static List<List<FileStatus>> getSplitStatus(JsonNode tree,
+                                                 long maxSplitSize,
+                                                 List<String> sessionSetupSqls) throws SQLException, IOException {
         var catalogSchemaAndTables =
                 Transformations.getAllTablesOrPathsFromSelect(Transformations.getFirstStatementNode(tree), null, null);
 
@@ -35,7 +48,7 @@ public interface SplitPlanner {
         if (splitPlanner == null) {
             throw new SQLException("unsupported type : " + tableFunction);
         }
-        var fileStatuses = splitPlanner.pruneFiles(tree, maxSplitSize, Map.of());
+        var fileStatuses = splitPlanner.pruneFiles(tree, maxSplitSize, Map.of(), sessionSetupSqls);
         //fileStatuses.sort(Comparator.comparing(FileStatus::lastModified));
         return getSplitStatus(maxSplitSize, fileStatuses);
     }
@@ -133,7 +146,16 @@ public interface SplitPlanner {
 
     static List<TreeAndSize> getSplitTreeAndSize(JsonNode tree,
                                                  long maxSplitSize) throws SQLException, IOException {
-        var splits = getSplitStatus(tree, maxSplitSize);
+        return getSplitTreeAndSize(tree, maxSplitSize, List.of());
+    }
+
+    /**
+     * @param sessionSetupSqls see {@link #getSplitStatus(JsonNode, long, List)}
+     */
+    static List<TreeAndSize> getSplitTreeAndSize(JsonNode tree,
+                                                 long maxSplitSize,
+                                                 List<String> sessionSetupSqls) throws SQLException, IOException {
+        var splits = getSplitStatus(tree, maxSplitSize, sessionSetupSqls);
         return splits.stream().map(split -> {
             var copy = tree.deepCopy();
             SplitPlanner.replacePathInFromClause(copy, split.stream().map(FileStatus::fileName).toArray(String[]::new));
diff --git a/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java b/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java
new file mode 100644
index 00000000..b3ee1093
--- /dev/null
+++ b/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java
@@ -0,0 +1,49 @@
+package io.dazzleduck.sql.commons.planner;
+
+import io.dazzleduck.sql.commons.Transformations;
+import io.dazzleduck.sql.commons.authorization.SessionVariables;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.sql.SQLException;
+import java.util.List;
+
+import static io.dazzleduck.sql.commons.util.TestConstants.SUPPORTED_HIVE_PATH_QUERY;
+
+/**
+ * Split planning prunes partitions on connections it opens itself, and the tree it prunes with
+ * already carries the injected row-level-security filter. A filter referencing
+ * {@code getvariable('d')} therefore evaluates against NULL — pruning away every file — unless the
+ * request's session variables are applied to those connections too.
+ */
+public class SplitPlannerSessionVariableTest {
+
+    /** example/data/hive_table has dt=2024-01-01 (1 file) and dt=2025-01-01 (2 files). */
+    private static final String QUERY_FILTERED_BY_VARIABLE =
+            SUPPORTED_HIVE_PATH_QUERY + " WHERE dt = getvariable('d')::DATE";
+
+    @Test
+    public void sessionVariableIsAppliedWhenPruningPartitions() throws SQLException, IOException {
+        var setupSqls = SessionVariables.toSetStatements("{\"d\":\"2025-01-01\"}");
+        Assertions.assertEquals(List.of("SET VARIABLE d = '2025-01-01'"), setupSqls);
+
+        var statuses = SplitPlanner.getSplitStatus(
+                Transformations.parseToTree(QUERY_FILTERED_BY_VARIABLE), 1024 * 1024 * 1024, setupSqls);
+
+        var files = statuses.stream().flatMap(List::stream).toList();
+        Assertions.assertEquals(2, files.size(),
+                "expected the two dt=2025-01-01 files; 0 means the filter was evaluated with the "
+                        + "variable unset (getvariable -> NULL) and every file was pruned");
+        Assertions.assertTrue(files.stream().allMatch(f -> f.fileName().contains("dt=2025-01-01")), files.toString());
+    }
+
+    @Test
+    public void withoutTheSessionVariableEveryFileIsPruned() throws SQLException, IOException {
+        // Documents the failure mode the parameter exists to prevent: same query, no variables.
+        var statuses = SplitPlanner.getSplitStatus(
+                Transformations.parseToTree(QUERY_FILTERED_BY_VARIABLE), 1024 * 1024 * 1024, List.of());
+
+        Assertions.assertEquals(0, statuses.stream().flatMap(List::stream).count());
+    }
+}
diff --git a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java
index c0c9382b..d5e8d28e 100644
--- a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java
+++ b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java
@@ -1405,8 +1405,7 @@ public class DuckDBFlightSqlProducer implements FlightSqlHttpProducer, SqlProduc
         // headers, so they cannot be overridden per-request. Applied as SET VARIABLE so queries and
         // injected RLS filters can read them via getvariable('name'). A malformed claim throws here
         // (before the connection is built) rather than failing silently.
-        sqls.addAll(SessionVariables.toSetStatements(
-                getVerifiedClaims(context).get(Headers.CLAIM_SESSION_VARIABLES)));
+        sqls.addAll(sessionSetupSqls(context));
         try {
             return ConnectionPool.getConnection(sqls.toArray(new String[0]));
         } catch (Exception e ){
@@ -1414,6 +1413,20 @@ public class DuckDBFlightSqlProducer implements FlightSqlHttpProducer, SqlProduc
         }
     }
 
+    /**
+     * The {@code SET VARIABLE} statements for this request's session variables.
+     *
+     * <p>{@link #getConnection} is not the only connection that evaluates the request's query: split
+     * planning prunes partitions on its own connections, and the tree it prunes with already has the
+     * row-level-security filter injected. A filter referencing {@code getvariable('x')} evaluates
+     * against NULL on a connection that has not run these — pruning away every file and returning an
+     * empty result — so every such connection must apply them too.
+     */
+    protected static List<String> sessionSetupSqls(CallContext context) {
+        return SessionVariables.toSetStatements(
+                getVerifiedClaims(context).get(Headers.CLAIM_SESSION_VARIABLES));
+    }
+
     protected static DatabaseSchema getDatabaseSchema(CallContext context, AccessMode accessMode){
         var verifiedClaims = getVerifiedClaims(context);
         if (accessMode == AccessMode.RESTRICTED) {
diff --git a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/RestrictedFlightSqlProducer.java b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/RestrictedFlightSqlProducer.java
index b3338149..a0629466 100644
--- a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/RestrictedFlightSqlProducer.java
+++ b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/RestrictedFlightSqlProducer.java
@@ -535,7 +535,10 @@ public class RestrictedFlightSqlProducer extends DuckDBFlightSqlProducer {
                                                         final FlightDescriptor descriptor) {
         try {
             var splitSize = getSplitSize(context);
-            var splits = SplitPlanner.getSplitTreeAndSize(tree, splitSize);
+            // The tree already carries the injected RLS filter; the pruners open their own
+            // connections, so they need this request's session variables or a getvariable(...)
+            // filter prunes against NULL and every file is discarded.
+            var splits = SplitPlanner.getSplitTreeAndSize(tree, splitSize, sessionSetupSqls(context));
             var list = splits.stream().map(split -> {
                 try {
                     var sql = Transformations.parseToSql(split.tree());
-- 
2.50.1 (Apple Git-155)

```

</details>

gtaneja-sentrinox:
## Review, part 2: the remaining findings

Follow-up to the split-planning fix above. One retraction first, then three real defects and some documentation.

### Retracting what I said about the trust model

In the earlier comment I described `x-dd-variables` being client-authored as a weakness. That was wrong, and the author has corrected me: the bundled `LoginService` is a **demo**, and in production the login service issues the token with the variables it has decided on. The plumbing for that is already right — `HttpCredentialValidator.validate` forwards every `claims.generate.headers` entry from the client's call headers to the configured `login_url` as the `claims` map, so a client's *requested* variables reach the login service, which decides what to sign. `LoginRequest.claims` is `Map<String,String>`, so the claim travels as JSON text throughout, consistent with `JwtClaimsExtractor` reading it as a `String`. Nothing to fix.

What does need fixing is that `CLAUDE.md` names the wrong enforcement point — it says `SessionVariables.validate(...)` is "the server-side place to enforce a variable policy". Per the design above that is the login service's job; `validate()` is defence-in-depth on the query server, and is an empty placeholder. The patch rewrites that section to describe the actual contract and to say plainly that the bundled `LoginService` is a demo.

### 1. `VALID_NAME` accepts identifiers DuckDB rejects — MEDIUM

`^[A-Za-z_][A-Za-z0-9_]*$` permits reserved words, and the name is rendered bare. Verified against DuckDB 1.5.4:

```
SET VARIABLE table  = 'x'   ->  Parser Error: syntax error at or near "table"
SET VARIABLE select = 'x'   ->  Parser Error
SET VARIABLE order  = 'x'   ->  Parser Error
SET VARIABLE from   = 'x'   ->  Parser Error
SET VARIABLE "table" = 'x'  ->  OK, getvariable('table') = 'x'
```

A token with `x-dd-variables = {"table":"x"}` fails every request from that client. Fixed by rendering the name as a quoted identifier — the regex already excludes the double quote, so nothing needs escaping.

### 2. A control character in a value fails the request confusingly — MEDIUM

DuckDB truncates a SQL string at an embedded NUL, so the rendered statement becomes an unterminated literal:

```
SET VARIABLE v = 'x<NUL>y'   ->  Parser Error: unterminated quoted string
```

**Not** an injection — every truncation point leaves the literal unterminated, because quote doubling always produces balanced pairs — but the request fails with an error naming neither the claim nor the variable. Now rejected up front with a message that names both.

Combined with the `ConnectionPool` change in the previous commit, this also closes the leak: before, a failed setup batch left the connection open, so a value with a NUL leaked a DuckDB connection on every request.

### 3. No bound on claim size — LOW

A claim could carry any number of variables of any length, and `getConnection` runs per request *and* per phase, so each was replayed several times. Capped at 64 variables and 4096 characters per value.

### Also

- `getConnection` reported any setup-batch failure as `NoSuchCatalogSchemaError` with the cause discarded. Only the `USE` can mean a missing schema. Now keeps the cause.
- The `Headers.java` comment says `x-dd-variables` is safe because it is "intentionally NOT in `SUPPORTED_HEADERS`". That set is referenced nowhere in the repo — only its own declaration — so it gates nothing. The real guarantee is that the query path resolves the claim from the signed token; the comment now says that instead.
- **Not changed, worth a look:** `dazzleduck-sql-login/reference.conf` adds `x-dd-variables` to `claims.generate.headers`, but nothing in `dazzleduck-sql-login` reads `JWT_TOKEN_CLAIMS_GENERATE_HEADERS_KEY` — `LoginService` uses `loginRequest.claims()` directly. That line reads as though the login service filters by the list when it does not.

### Tests

`SessionVariablesTest` goes 11 → 15. The important change is that the escaping is now **executed**, not just string-matched. DuckDB JDBC does run multiple statements from one `execute()` — I confirmed `SELECT 1; INSERT INTO t VALUES (99)` inserts the row — so quote doubling is the only thing between the claim and arbitrary SQL. The new test runs each payload and checks a canary table stays empty:

```java
setup.execute("CREATE OR REPLACE TEMP TABLE injection_canary(x INT)");
// ... render + execute each payload, assert the value round-trips verbatim ...
assertEquals(0, rs.getInt(1), "no payload may execute as SQL");
```

Reserved-word names are executed too, rather than only compared against an expected string.

JDK 21 on `4ebf72fd`: commons 502, flight 151, http 109 — 0 failures (skips are the pre-existing Testcontainers ones).

### Patch

Applies on top of the split-planning patch in the previous comment.

<details>
<summary><code>0002-fix-auth-harden-session-variable-rendering.patch</code></summary>


> **Patch re-posted.** The first upload of this patch was corrupted in transit: the six
> characters of a unicode escape in one test were normalised to `^@`, which applied cleanly
> but made that test fail at JSON parsing instead of the check it was written for. The test
> now builds the character in Java instead, and this patch has been verified by extracting it
> back out of this comment, applying it with `git am`, and running the suites.

`````diff
From eb970d257d5e02d9804df945df9c58e87ffec6f4 Mon Sep 17 00:00:00 2001
From: Gagan Taneja <gtaneja@sentrinox.ai>
Date: Mon, 14 Sep 2026 12:22:53 -0700
Subject: [PATCH] fix(auth): harden session-variable rendering and clarify who
 owns the policy
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

Three defects in how a x-dd-variables claim becomes SQL, plus documentation
that pointed at the wrong enforcement point.

Reserved words. VALID_NAME accepts any [A-Za-z_][A-Za-z0-9_]* identifier, which
includes table, select, order, from, group and default — all of which DuckDB
rejects unquoted, so `SET VARIABLE table = 'x'` is a parser error. Render the
name as a quoted identifier; the regex already excludes the double quote, so
there is nothing to escape.

Control characters. DuckDB truncates a SQL string at an embedded NUL, turning
the rendered statement into an unterminated literal and failing the request with
a parser error that names neither the claim nor the variable. Reject control
characters while the message can still be useful. (Not an injection: every
truncation point leaves the literal unterminated, because quote doubling always
produces balanced pairs.)

Unbounded size. A claim could carry any number of variables of any length, and
getConnection runs per request and per phase, so each one was replayed several
times. Cap at 64 variables and 4096 characters per value.

Also stop reporting a failed setup batch as NoSuchCatalogSchemaError with the
cause discarded — only the USE can mean a missing schema, and a failed
SET VARIABLE sent the caller looking in the wrong place.

Documentation: the trust model said SessionVariables.validate(...) was "the
server-side place to enforce a variable policy". It is not. A client *requests*
variables, HttpCredentialValidator forwards every claims.generate.headers entry
to the configured login_url, and the login service decides what to set, override
or reject before signing. Say that, note that the bundled LoginService is a demo
that signs whatever it is given, and demote validate() to the defence-in-depth
hook it actually is. The Headers comment claimed SUPPORTED_HEADERS provided the
per-request protection; that set is referenced nowhere, and the real guarantee
is that the query path resolves the claim from the signed token.

Tests: reserved-word names now execute against DuckDB rather than only matching
an expected string, control characters and size caps are covered, and a canary
table proves the injection payloads render as data — DuckDB JDBC will execute
multiple statements in one execute(), so that escaping is load-bearing.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
---
 CLAUDE.md                                     |  39 +++++--
 .../io/dazzleduck/sql/common/Headers.java     |  14 ++-
 .../authorization/SessionVariables.java       |  61 +++++++++-
 .../authorization/SessionVariablesTest.java   | 106 +++++++++++++++++-
 .../SplitPlannerSessionVariableTest.java      |   2 +-
 .../server/DuckDBFlightSqlProducer.java       |   5 +-
 .../server/NoSuchCatalogSchemaError.java      |   9 ++
 7 files changed, 209 insertions(+), 27 deletions(-)

diff --git a/CLAUDE.md b/CLAUDE.md
index db75998c..ff1cf956 100644
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,18 +152,33 @@ verified (signed) token **only** — it is intentionally not a recognized reques
 client cannot override it per request. Every value must be a **quoted JSON string** — a bare
 number or boolean (`{"n":42}`) is rejected with a hint to quote it, since all variables are
 applied as VARCHAR literals; cast for numeric/temporal comparisons (`getvariable('n')::INT`).
-Variable names must match `[A-Za-z_][A-Za-z0-9_]*`; a malformed claim fails the request.
-
-**Trust model.** "Verified-claim only" means the value cannot be overridden per request on an
-already-issued token — the claim is not a recognized request header, so a query request cannot
-change it. It does **not** mean the value is server-authored: the claim is populated at token
-issuance, and in the Flight basic-auth path `AdvanceJWTTokenAuthenticator` copies every
-`claims.generate.headers` entry from the connection headers into the signed token — so a Flight
-client can set `x-dd-variables` via its connection string, exactly as it can for `x-dd-access` /
-`x-dd-filter`. Assigning variables from the authenticated identity (e.g. a fixed tenant per user)
-rather than trusting client-supplied values is the login/auth service's responsibility. The
-`SessionVariables.validate(...)` hook is the server-side place to enforce a variable policy
-(allowed names, value constraints) before the variables are applied.
+Variable names must match `[A-Za-z_][A-Za-z0-9_]*` and are rendered as quoted identifiers, so a
+reserved word such as `table` works. Values may not contain control characters, are capped at 4096
+characters, and at most 64 variables may be sent; a malformed claim fails the request.
+
+The claim value is the JSON **text**, i.e. a claim whose value is a string. A token minted with
+`x-dd-variables` as a nested JSON object is rejected at parse time, because claims are read as
+strings.
+
+**Trust model — the login service owns the policy.** "Verified-claim only" means the value cannot
+be overridden per request on an already-issued token: the query path resolves it from the signed
+claims, never from a request header. It does **not** mean the query server chose the value. The
+value is decided at token issuance, and deciding it is the **login service's** job:
+
+- A client *requests* variables (a Flight connection property / call header, or the `claims` map in
+  a `POST /v1/login` body).
+- `HttpCredentialValidator` forwards every `claims.generate.headers` entry — `x-dd-variables`
+  included — to the configured `login_url` as the `claims` map.
+- The login service decides what to set, override, or reject, and signs only what it approves. A
+  production one assigns variables from the authenticated identity (e.g. a fixed tenant per user).
+
+The bundled `LoginService` is a **demo**: it signs `loginRequest.claims()` as given, with no policy.
+Do not run it in production. Likewise, when the Flight server mints tokens itself
+(`jwt_token.generation = true` with no `login_url`), `AdvanceJWTTokenAuthenticator` copies the
+connection headers into the token unfiltered — also a development mode.
+
+`SessionVariables.validate(...)` on the query server is a defence-in-depth hook (allowed names,
+value constraints), not the primary control; it is an unimplemented placeholder today.
 
 **External access control** (for restricted modes):
 ```sql
diff --git a/dazzleduck-sql-common/src/main/java/io/dazzleduck/sql/common/Headers.java b/dazzleduck-sql-common/src/main/java/io/dazzleduck/sql/common/Headers.java
index 432e8d2d..d59f7fde 100644
--- a/dazzleduck-sql-common/src/main/java/io/dazzleduck/sql/common/Headers.java
+++ b/dazzleduck-sql-common/src/main/java/io/dazzleduck/sql/common/Headers.java
@@ -36,9 +36,17 @@ public class Headers {
     public static final String HEADER_ACCESS = "x-dd-access";
     public static final String HEADER_ACCESS_TYPE = "x-dd-access-type";
     // Session variables applied to the DuckDB connection as SET VARIABLE, referenceable in SQL
-    // and RLS filters via getvariable('name'). JSON object of string key/values, e.g.
-    // {"tenant_id":"acme"}. Trusted from the verified JWT claim ONLY — intentionally NOT in
-    // SUPPORTED_HEADERS, so a client-supplied request header of the same name is never honored.
+    // and RLS filters via getvariable('name'). The claim value is the JSON *text* of an object of
+    // string key/values, e.g. the string {"tenant_id":"acme"}.
+    //
+    // Read from the verified JWT claim only: a query request cannot change it, because the query
+    // path resolves it from the signed claims (see DuckDBFlightSqlProducer.sessionSetupSqls) and
+    // never from the request headers. Note this is NOT enforced by SUPPORTED_HEADERS, which is
+    // referenced nowhere.
+    //
+    // The value is chosen at token issuance, not by this server: HttpCredentialValidator forwards
+    // every claims.generate.headers entry to the configured login service, which decides what to
+    // sign. Deciding which variables a given identity may set is that service's job.
     public static final String CLAIM_SESSION_VARIABLES = "x-dd-variables";
     public static final String HEADER_SPLIT_SIZE = "x-dd-split-size";
     public static final String HEADER_DATA_PARTITION = "x-dd-partition";
diff --git a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/authorization/SessionVariables.java b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/authorization/SessionVariables.java
index 59aa011e..0cb31f78 100644
--- a/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/authorization/SessionVariables.java
+++ b/dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/authorization/SessionVariables.java
@@ -29,9 +29,16 @@ public final class SessionVariables {
 
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
-    /** DuckDB variable name we allow to appear unquoted in {@code SET VARIABLE}. */
+    /**
+     * Variable name we accept. Deliberately excludes the double quote, so the name can be rendered
+     * as a quoted identifier without any escaping of its own.
+     */
     private static final Pattern VALID_NAME = Pattern.compile("^[A-Za-z_][A-Za-z0-9_]*$");
 
+    /** Upper bounds so an oversized claim cannot turn one request into thousands of statements. */
+    private static final int MAX_VARIABLES = 64;
+    private static final int MAX_VALUE_LENGTH = 4096;
+
     private SessionVariables() {
     }
 
@@ -40,8 +47,14 @@ public final class SessionVariables {
      * {@code claimJson}. Returns an empty list when the claim is null, blank, or an empty object.
      * Null-valued entries are skipped ({@code getvariable} of an unset name already returns NULL).
      *
-     * @throws IllegalArgumentException if the claim is not a JSON object of scalar values, or a
-     *                                  key is not a valid identifier
+     * <p>The claim value is the JSON <em>text</em>, i.e. a JWT claim whose value is a string such as
+     * {@code "{\"tenant_id\":\"acme\"}"} — not a nested JSON object, which
+     * {@code JwtClaimsExtractor} cannot read as a String.
+     *
+     * @throws IllegalArgumentException if the claim is not a JSON object of scalar values, a key is
+     *                                  not a valid identifier, a value holds a control character,
+     *                                  or the claim exceeds {@link #MAX_VARIABLES} entries or
+     *                                  {@link #MAX_VALUE_LENGTH} characters in a value
      */
     public static List<String> toSetStatements(String claimJson) {
         if (claimJson == null || claimJson.isBlank()) {
@@ -78,14 +91,32 @@ public final class SessionVariables {
                         "Session variable '" + name + "': " + value + " needs to be inside quotes "
                                 + "(write it as \"" + name + "\": \"" + value + "\")");
             }
-            variables.put(name, (String) value);
+            String text = (String) value;
+            if (text.length() > MAX_VALUE_LENGTH) {
+                throw new IllegalArgumentException("Session variable '" + name + "' is too long: "
+                        + text.length() + " characters, limit is " + MAX_VALUE_LENGTH);
+            }
+            // DuckDB truncates a SQL string at an embedded NUL, which turns the rendered statement
+            // into an unterminated literal and fails the whole request with an unrelated parser
+            // error. Reject control characters here, while the message can still name the variable.
+            int control = indexOfControlCharacter(text);
+            if (control >= 0) {
+                throw new IllegalArgumentException(String.format(
+                        "Session variable '%s' contains a control character (U+%04X) at index %d",
+                        name, (int) text.charAt(control), control));
+            }
+            variables.put(name, text);
+        }
+        if (variables.size() > MAX_VARIABLES) {
+            throw new IllegalArgumentException("Too many session variables: " + variables.size()
+                    + ", limit is " + MAX_VARIABLES);
         }
 
         validate(variables);
 
         List<String> sqls = new ArrayList<>(variables.size());
         variables.forEach((name, value) ->
-                sqls.add("SET VARIABLE " + name + " = " + sqlLiteral(value)));
+                sqls.add("SET VARIABLE " + quotedIdentifier(name) + " = " + sqlLiteral(value)));
         return sqls;
     }
 
@@ -106,6 +137,26 @@ public final class SessionVariables {
         // rules). No additional checks yet.
     }
 
+    /** Index of the first ISO control character in {@code value}, or -1 if there is none. */
+    private static int indexOfControlCharacter(String value) {
+        for (int i = 0; i < value.length(); i++) {
+            if (Character.isISOControl(value.charAt(i))) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    /**
+     * A double-quoted SQL identifier. {@code VALID_NAME} already rejects the double quote, so there
+     * is nothing to escape. Quoting matters because the regex accepts reserved words — bare
+     * {@code SET VARIABLE table = 'x'} is a DuckDB parser error, {@code SET VARIABLE "table" = 'x'}
+     * is not.
+     */
+    private static String quotedIdentifier(String name) {
+        return "\"" + name + "\"";
+    }
+
     /** A single-quoted SQL string literal with embedded quotes doubled. */
     private static String sqlLiteral(String value) {
         return "'" + value.replace("'", "''") + "'";
diff --git a/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/authorization/SessionVariablesTest.java b/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/authorization/SessionVariablesTest.java
index 4303cb09..883bbd95 100644
--- a/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/authorization/SessionVariablesTest.java
+++ b/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/authorization/SessionVariablesTest.java
@@ -1,7 +1,10 @@
 package io.dazzleduck.sql.commons.authorization;
 
+import io.dazzleduck.sql.commons.ConnectionPool;
 import org.junit.jupiter.api.Test;
 
+import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -15,8 +18,8 @@ class SessionVariablesTest {
         List<String> sqls = SessionVariables.toSetStatements(
                 "{\"tenant_id\":\"acme\",\"region\":\"us-east\"}");
         assertEquals(List.of(
-                "SET VARIABLE tenant_id = 'acme'",
-                "SET VARIABLE region = 'us-east'"), sqls);
+                "SET VARIABLE \"tenant_id\" = 'acme'",
+                "SET VARIABLE \"region\" = 'us-east'"), sqls);
     }
 
     @Test
@@ -43,7 +46,7 @@ class SessionVariablesTest {
 
     @Test
     void quotedNumberIsAcceptedAsAString() {
-        assertEquals(List.of("SET VARIABLE var1 = '42'"),
+        assertEquals(List.of("SET VARIABLE \"var1\" = '42'"),
                 SessionVariables.toSetStatements("{\"var1\":\"42\"}"));
     }
 
@@ -55,7 +58,7 @@ class SessionVariablesTest {
     @Test
     void singleQuotesInValueAreDoubled() {
         List<String> sqls = SessionVariables.toSetStatements("{\"name\":\"O'Brien\"}");
-        assertEquals(List.of("SET VARIABLE name = 'O''Brien'"), sqls);
+        assertEquals(List.of("SET VARIABLE \"name\" = 'O''Brien'"), sqls);
     }
 
     @Test
@@ -63,7 +66,7 @@ class SessionVariablesTest {
         // A crafted value must stay entirely inside one quoted literal, never break out into SQL.
         List<String> sqls = SessionVariables.toSetStatements(
                 "{\"tenant_id\":\"x'; DROP TABLE orders; --\"}");
-        assertEquals(List.of("SET VARIABLE tenant_id = 'x''; DROP TABLE orders; --'"), sqls);
+        assertEquals(List.of("SET VARIABLE \"tenant_id\" = 'x''; DROP TABLE orders; --'"), sqls);
     }
 
     @Test
@@ -91,4 +94,97 @@ class SessionVariablesTest {
         assertThrows(IllegalArgumentException.class,
                 () -> SessionVariables.toSetStatements("[\"a\",\"b\"]"));
     }
+
+    @Test
+    void reservedWordNamesAreRenderedAsQuotedIdentifiers() throws Exception {
+        // The name regex accepts reserved words, so the identifier has to be quoted: bare
+        // `SET VARIABLE table = 'x'` is a DuckDB parser error.
+        for (String reserved : List.of("table", "select", "order", "from", "group", "default")) {
+            var sqls = SessionVariables.toSetStatements("{\"" + reserved + "\":\"x\"}");
+            assertEquals(List.of("SET VARIABLE \"" + reserved + "\" = 'x'"), sqls);
+            assertExecutable(sqls, reserved, "x");
+        }
+    }
+
+    @Test
+    void controlCharacterInValueIsRejected() throws Exception {
+        // DuckDB truncates a SQL string at an embedded NUL, which would turn the rendered statement
+        // into an unterminated literal and fail the request with an unrelated parser error.
+        // Build the NUL from a char rather than writing a unicode escape in the source: that
+        // escape sequence does not survive being copied through tools that normalise it.
+        var ex = assertThrows(IllegalArgumentException.class,
+                () -> SessionVariables.toSetStatements(claimWithValue("x" + (char) 0 + "y")));
+        assertTrue(ex.getMessage().contains("control character"), ex.getMessage());
+        assertThrows(IllegalArgumentException.class,
+                () -> SessionVariables.toSetStatements(claimWithValue("line1" + (char) 10 + "line2")));
+    }
+
+    /** {@code {"a":"<value>"}}, with value escaped by Jackson rather than written by hand. */
+    private static String claimWithValue(String value) throws Exception {
+        return new com.fasterxml.jackson.databind.ObjectMapper()
+                .writeValueAsString(java.util.Map.of("a", value));
+    }
+
+    @Test
+    void oversizedClaimsAreRejected() {
+        var manyVars = new StringBuilder("{");
+        for (int i = 0; i < 65; i++) {
+            manyVars.append(i > 0 ? "," : "").append("\"v").append(i).append("\":\"x\"");
+        }
+        var tooMany = assertThrows(IllegalArgumentException.class,
+                () -> SessionVariables.toSetStatements(manyVars.append("}").toString()));
+        assertTrue(tooMany.getMessage().contains("Too many session variables"), tooMany.getMessage());
+
+        var tooLong = assertThrows(IllegalArgumentException.class,
+                () -> SessionVariables.toSetStatements("{\"a\":\"" + "x".repeat(4097) + "\"}"));
+        assertTrue(tooLong.getMessage().contains("too long"), tooLong.getMessage());
+    }
+
+    @Test
+    void renderedStatementsRunOnDuckDbAndCannotInjectSql() throws Exception {
+        // Asserting the rendered string is not the same as proving it is safe: DuckDB JDBC executes
+        // multiple statements in one execute(), so the escaping is what stands between the claim and
+        // arbitrary SQL. Run the payloads and check nothing but the variable changed.
+        try (var connection = ConnectionPool.getConnection();
+             Statement setup = connection.createStatement()) {
+            setup.execute("CREATE OR REPLACE TEMP TABLE injection_canary(x INT)");
+            for (String payload : List.of("x'; INSERT INTO injection_canary VALUES (1); --",
+                    "back\\slash'; INSERT INTO injection_canary VALUES (2); --",
+                    "O'Brien",
+                    "a' || (SELECT 'b') || '")) {
+                var sqls = SessionVariables.toSetStatements(
+                        "{\"v\":" + new com.fasterxml.jackson.databind.ObjectMapper().writeValueAsString(payload) + "}");
+                for (String sql : sqls) {
+                    try (Statement st = connection.createStatement()) {
+                        st.execute(sql);
+                    }
+                }
+                try (Statement st = connection.createStatement();
+                     var rs = st.executeQuery("SELECT getvariable('v')")) {
+                    assertTrue(rs.next());
+                    assertEquals(payload, rs.getString(1), "value must round-trip verbatim");
+                }
+            }
+            try (Statement st = connection.createStatement();
+                 var rs = st.executeQuery("SELECT count(*) FROM injection_canary")) {
+                assertTrue(rs.next());
+                assertEquals(0, rs.getInt(1), "no payload may execute as SQL");
+            }
+        }
+    }
+
+    private static void assertExecutable(List<String> sqls, String name, String expected) throws SQLException {
+        try (var connection = ConnectionPool.getConnection()) {
+            for (String sql : sqls) {
+                try (Statement st = connection.createStatement()) {
+                    st.execute(sql);
+                }
+            }
+            try (Statement st = connection.createStatement();
+                 var rs = st.executeQuery("SELECT getvariable('" + name + "')")) {
+                assertTrue(rs.next());
+                assertEquals(expected, rs.getString(1));
+            }
+        }
+    }
 }
diff --git a/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java b/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java
index b3ee1093..ec6a08f4 100644
--- a/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java
+++ b/dazzleduck-sql-commons/src/test/java/io/dazzleduck/sql/commons/planner/SplitPlannerSessionVariableTest.java
@@ -26,7 +26,7 @@ public class SplitPlannerSessionVariableTest {
     @Test
     public void sessionVariableIsAppliedWhenPruningPartitions() throws SQLException, IOException {
         var setupSqls = SessionVariables.toSetStatements("{\"d\":\"2025-01-01\"}");
-        Assertions.assertEquals(List.of("SET VARIABLE d = '2025-01-01'"), setupSqls);
+        Assertions.assertEquals(List.of("SET VARIABLE \"d\" = '2025-01-01'"), setupSqls);
 
         var statuses = SplitPlanner.getSplitStatus(
                 Transformations.parseToTree(QUERY_FILTERED_BY_VARIABLE), 1024 * 1024 * 1024, setupSqls);
diff --git a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java
index d5e8d28e..43af8bbb 100644
--- a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java
+++ b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/DuckDBFlightSqlProducer.java
@@ -1409,7 +1409,10 @@ public class DuckDBFlightSqlProducer implements FlightSqlHttpProducer, SqlProduc
         try {
             return ConnectionPool.getConnection(sqls.toArray(new String[0]));
         } catch (Exception e ){
-            throw new NoSuchCatalogSchemaError(dbSchema);
+            // Only the USE can mean "no such catalog/schema". The batch also carries the session
+            // variables now, and reporting a failed SET VARIABLE as a missing schema sends the
+            // caller looking in the wrong place.
+            throw new NoSuchCatalogSchemaError(dbSchema, e);
         }
     }
 
diff --git a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/NoSuchCatalogSchemaError.java b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/NoSuchCatalogSchemaError.java
index 42113dc0..41f71396 100644
--- a/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/NoSuchCatalogSchemaError.java
+++ b/dazzleduck-sql-flight/src/main/java/io/dazzleduck/sql/flight/server/NoSuchCatalogSchemaError.java
@@ -6,4 +6,13 @@ public class NoSuchCatalogSchemaError extends Exception {
         super(String.format("Catalog or Schema %s Not Fount", catalogAndSchema));
         this.catalogAndSchema = catalogAndSchema;
     }
+
+    /**
+     * The connection setup batch carries more than the USE — a failed SET VARIABLE surfaces here
+     * too. Keep the real cause so the logs say which statement actually failed.
+     */
+    public NoSuchCatalogSchemaError(String catalogAndSchema, Throwable cause) {
+        super(String.format("Catalog or Schema %s Not Fount", catalogAndSchema), cause);
+        this.catalogAndSchema = catalogAndSchema;
+    }
 }
-- 
2.50.1 (Apple Git-155)


`````

</details>

PR #417, opened 2026-09-15
fix(transform): prune a shared RLS filter-CTE per consumer, not per outer scope

gtaneja-sentrinox:
`injectFilterCtes` keys `tablesToWrap` by qualified table name, so **one** filter CTE is emitted per table. A query naming the same view twice —

```sql
WITH page AS (SELECT rowid, ts FROM v ORDER BY ts DESC LIMIT 200)
SELECT <cols> FROM v JOIN page p ON p.rowid = v.rowid
```

— has both references rewritten to a single `___v`, which makes that CTE's consumers a **set** rather than one scope.

## The bug

`pruneStarFilterCte` recovers consumer usage from the enclosing scope **minus its own CTE bodies**:

```java
UsedColumns consumer = collectScopedUsage(consumerScope, /*excludeCteMap=*/true);
```

That exclusion is deliberate and correct while there is one consumer. With two it drops columns a sibling CTE still reads, the pruned view body no longer projects them, and **the query stops binding**.

`sharedFilterCte_columnUsedOnlyInSiblingCte_stillBinds` reproduces it — on `main` it is the only red test in the class.

## The fix

Collecting the union of all consumers would fix the binding but not the cost: every join any consumer needs would be kept for all of them, so a two-column consumer inherits a wide one's joins.

So the CTE is **split** — one clone per consumer, each pruned to that scope's own usage, each consumer's references rewritten to its own clone.

- Clones are inserted at the original's index, so they stay declared ahead of every scope that reads them.
- The split is abandoned wholesale if any consumer cannot be rewritten.
- A consumer whose columns are not enumerable (a STAR in that scope) gets an **unpruned** clone rather than blocking the others.
- A **single** consumer still delegates to the original path, bit-for-bit unchanged.

Splitting duplicates the filter expression, never the rows: each clone is the identical `SELECT * FROM <view> WHERE <filter>` over a different subset of the view's joins, so every consumer still reads exactly the rows the grant allows.

## Why this is wanted

It unblocks a two-phase page materialization downstream: pick the page on a two-column projection, then join back to the view to dress only those rows. Measured 2.1–3.3× on a synthetic DuckLake, flat across file layouts, and it is the only lever that reaches a `SELECT *` request — which today gets no pruning at all, because a star projection bails out of the transform. Without per-consumer specialization the narrow phase inherits the wide phase's joins and the optimization buys nothing.

## Testing

`./mvnw -pl dazzleduck-sql-commons test` → **505 tests, 0 failures**, `PruneUnusedLeftJoinsTest` at **53** (3 new):

| test | pins |
|---|---|
| `sharedFilterCte_columnUsedOnlyInSiblingCte_stillBinds` | the regression — red before this change |
| `sharedFilterCte_narrowSiblingAndWideOuter_bothCorrect` | the two-phase shape stays equivalent to the view |
| `sharedFilterCte_narrowConsumerGetsItsOwnJoinFreeClone` | the specialization: page's clone carries 0 joins, the outer's carries 1 |

All three assert `assertEquivalentToView`, so the rewrites are checked against the real view rather than only structurally.

> **JDK 21 is required to run the suite.** On JDK 25, Arrow's netty allocator fails to initialise and 35 pre-existing tests error out in `TestUtils.isEqual` before reaching any of this.


tanejagagan:
## Review

Core idea is correct and the tests are strong — all three new cases assert `assertEquivalentToView`, so the rewrites are checked against the real view, not just structurally. The concern is that `splitStarFilterCte` mutates shared AST state (`outerCopy` / `cte_map`) in ways the enclosing loop and its own bail-outs don't fully account for. **Three correctness issues, two minor.**

### 1. Stale `cte_map` indices after the split resizes the array (most severe)

`pruneUnusedLeftJoins` iterates a **precomputed** `List<Integer> cteRefs` and indexes `cte_map` by those values, but `splitStarFilterCte` does `map.remove(cteIndex)` then inserts N clones — changing the array size and shifting every later index.

With two star-filter CTEs over the same view, `cteRefs = [0, 1]`. Splitting index 0 (remove 1, insert 2 → net +1) turns the map into `[w1__c0, w1__c1, w2, page]`. The loop's next `map.get(1)` then returns the clone `w1__c1` instead of `w2` — `w2` is never processed and a clone gets re-fed into the transform, producing a mis-pruned or corrupt AST.

Reachability is gated on ≥2 star-filter CTEs over the same view in one call (uncommon, since the authorizer emits one per table), but indexing a precomputed list into an array the loop body structurally mutates is unambiguously wrong. Prefer iterating the split-eligible CTE by identity, or rebuilding the map once at the end.

### 2. Renames applied before the `!anyPruned` bail are never rolled back

`splitStarFilterCte` rewrites each consumer's `___fv` refs to `___fv__c{i}` inside the loop, then bails at `if (!anyPruned) return false;` **after** the renames, without inserting the clones. Consumers now reference CTE names that were never declared.

It's masked only when `changed` stays false for the whole call (the mutated `rootCopy` is discarded and the original AST returned). But if any other eligible reference in the same `pruneUnusedLeftJoins` invocation prunes successfully, the corrupt `rootCopy` is returned and the query fails to bind on `___fv__c0`. Fail-closed (no data leak), but a broken query. Fix: stage the renames + insert and only commit them after `anyPruned` is confirmed.

### 3. `consumers.size() <= 1` fast path ignores a sole *sibling-CTE* consumer

When the only consumer is a sibling CTE (the outer statement references `page`, not `___fv`), `findCteConsumers` returns `[page]`, size 1 → delegates to `pruneStarFilterCte(outerCopy, …)`, which collects usage from the outer scope **excluding** CTE bodies. `page`'s columns are invisible there, so they're pruned out of the inlined body and `page` no longer binds.

```sql
WITH ___fv AS (SELECT * FROM fv WHERE f_col > 0),
     page AS (SELECT a_name, f_id FROM ___fv)
SELECT p.f_id FROM page p
```

`a_name` (used only inside `page`) gets pruned away. This is a pre-existing latent bug (`main` takes the same path, not a regression from this PR), but it lives squarely in the multi-scope territory the PR set out to fix.

### Minor

4. **Clone name collision:** `cteName + "__c" + i` has no guard against an existing CTE of that literal name → a duplicate `cte_map` key. Negligible with `___`-prefixed authorizer-generated names, but unguarded.
5. **Redundant traversals:** each consumer's subtree is walked ~3× (existence check in `findCteConsumers`, `collectScopedUsage`, then `renameTableRefs`) — on the order of `3N` full recursive walks per RLS-wrapped query. Cost only, not correctness.

### Note

The `renameTableRefs(...) return false` "abandon the split entirely" branch appears unreachable: consumers are selected by the same `referencesTable` / `excludeCteMap` predicate that `countTableRefs` uses, so any scope found as a consumer always renames ≥1 ref. Harmless defensive code, but the comment describes a dead branch.

**Bottom line:** #1 and #2 share a root cause — `splitStarFilterCte` mutating shared state non-atomically. Both are fixable by validating fully (build clones, confirm `anyPruned`) before committing any mutation, and by not indexing a precomputed list into an array the loop body resizes.

tanejagagan:
Pushed `fe5a3a25` addressing the review:

- **#1 (stale `cte_map` indices):** the reference loop now visits high-index-first, so a split's `remove`+`insert` only ever shifts entries already processed.
- **#2 (unrolled renames on bail):** `splitStarFilterCte` now builds and prunes the clones off to the side and mutates the shared tree only after the split is guaranteed to complete — no half-rewritten AST on an abandoned split.
- **#3 (sole sibling-CTE consumer):** a single consumer is pruned against its own scope rather than the top-level statement, so a column read only inside a sibling CTE survives.
- **#4 (clone name collision):** clone keys go through `uniqueCteKey`.

Two regression tests added (`sharedFilterCte_soleConsumerIsSiblingCte_keepsItsColumns`, `twoFilterCtesOverSameView_secondStillPrunedAfterFirstSplits`) — both red before the commit. `PruneUnusedLeftJoinsTest` is at 55, full `dazzleduck-sql-commons` suite green on JDK 21 (the one error locally is `HivePartitionPruningTest` failing to pull its MinIO container — environmental, unrelated).

PR #427, opened 2026-09-16
feat(compactor): expose max_compacted_files for minor and major compaction

### LINKED ISSUE #425: Compaction: expose DuckLake max_compacted_files for minor and major — unbounded batch never commits on large lakes
## Summary

`CompactionService` never passes DuckLake's `max_compacted_files` parameter, so every compaction cycle attempts to merge **the entire eligible file set in a single transaction**. On a lake with a few thousand small files this reliably exhausts memory and/or temp space, never commits, and leaves orphaned Parquet objects behind.

## Where

`dazzleduck-sql-ducklake-compactor/.../CompactionService.java`, `runMinor()`:

```java
String sql = "CALL ducklake_merge_adjacent_files('%s', max_file_size := %d)"
        .formatted(database, config.minorCompactionMaxSize());
```

Only `max_file_size` is supplied. `max_compacted_files` is omitted, and DuckLake defaults it to effectively unbounded — from `ducklake_compaction_functions.cpp`:

```cpp
uint64_t max_files = NumericLimits<uint64_t>::Maximum() - 1;
auto max_files_entry = input.named_parameters.find("max_compacted_files");
```

The same applies to the major path via `MajorCompactor`.

## Why the existing knobs don't substitute for it

`min/major_compaction_max_size` is an **input filter**, not a batch limit — DuckLake applies it as:

```cpp
file_filter_clause += StringUtil::Format(" AND data.file_size_bytes < %llu", effective_max_file_size);
```

So when the average file is well below the threshold, *every* file is eligible regardless of how the threshold is tuned. Lowering it selects the same set; raising it adds nothing.

## Observed behaviour

Partitioned table, ~7,000 live files averaging ~875 kB (partition fan-out produces many small files per flush):

| Symptom | Value |
|---|---|
| Files eligible per cycle | ~7,000 (all of them) |
| Temp spill for one merge | **13 GB** |
| Cycles completed successfully | **0** over several hours |
| `totalFilesCompacted` | **0** |
| Backlog during observation | 6,935 → 10,079 files |

Each attempt writes merged Parquet to object storage, then fails before the catalog commit — so the objects are orphaned and the source files stay live, and the next cycle re-merges the identical set. It is a non-terminating loop that consumes storage and CPU while making no progress.

These were each tried and none helped, because none of them bounds the batch:

| Change | Result |
|---|---|
| `memory_limit` 6 → 8 → 10 → 12 GB | still fails |
| `threads` 4 → 2 | still fails |
| JVM `-Xmx` 4g → 2g | still fails |
| `minor/major_compaction_max_size` 96/256MB → 16/64MB | still fails |
| temp dir 3.2 GB → 100 GB | OOM resolved, still never commits |

## Requested change

Expose `max_compacted_files` as configuration for both minor and major compaction, e.g.:

```hocon
dazzleduck_sql_compaction {
  minor_compaction_max_files = 500
  major_compaction_max_files = 200
}
```

emitted as:

```sql
CALL ducklake_merge_adjacent_files('db', max_file_size := N, max_compacted_files := M)
```

A bounded batch commits in seconds, so each cycle makes durable incremental progress and a failure costs one small batch rather than the whole table. This also removes the need to raise memory, temp space and transaction timeouts to accommodate one enormous transaction.

Two other DuckLake parameters are also unused and may be worth exposing: `min_file_size` (skip files already large enough) and `newer_than`.

tanejagagan:
## Summary
- Fixes #425. `ducklake_merge_adjacent_files` never received `max_compacted_files`, so every compaction cycle attempted to merge the entire eligible file set in one transaction — on a large lake this exhausts memory/temp space and never commits, leaving orphaned Parquet files and a backlog that only grows.
- Adds `minor_max_compacted_files` (default `500`) and `major_max_compacted_files` (default `100`) as required config keys under `dazzleduck_sql_compaction`, passed through as `CALL ducklake_merge_adjacent_files(..., max_file_size := N, max_compacted_files := M)` for both the minor and major compaction paths. Separate knobs because minor runs far more often over many more small files than major.
- Startup now fails fast (`ConfigException.Missing`) if either key is absent from a custom config, rather than silently running uncapped.

## Test plan
- [x] `./mvnw test -pl dazzleduck-sql-ducklake-compactor -am` — all 19 tests pass (`CompactionConfigOverrideTest`, `CompactionIntegrationTest`, `CompactionOutcomeTest`)
- [x] Verified the DuckLake `ducklake_merge_adjacent_files` procedure signature accepts `max_compacted_files` as a `UBIGINT` named parameter via `duckdb_functions()`


tanejagagan:
Superseding this with #429, which covers the same max_compacted_files gap for minor compaction with a cleaner optional/0=unbounded design. Follow-up work (major-side equivalent, independent minor/major scheduling, per-cycle connection settings) will build on top of #429.

PR #430, opened 2026-09-16
feat(compactor): decouple minor/major scheduling, enable flags, connection settings

tanejagagan:
## Summary

**Stacked on #429** (not yet merged) — this diff includes #429's `minor_compaction_max_files` changes until that merges; it will shrink to just this PR's own changes afterward.

Two gaps in the current design:
1. **No way to disable just minor or major.** Worse, the "obvious" workaround — setting `minor_compaction_frequency` to something huge — also silently starves major, since today there's only ONE scheduled task per database (`runCompaction`) and major's eligibility is only ever checked when *that* tick fires, which minor's own frequency governs.
2. **No way to give minor and major different DuckDB connection settings** (e.g. `memory_limit`, `threads`) — both used a bare `ConnectionPool.getConnection()`.

### Changes
- `minor_compaction.enabled` / `major_compaction.enabled` (default `true`) — each now runs on its own independent `scheduleAtFixedRate` task instead of sharing one.
- Adds `major_compaction_max_files` (optional, `0`=unbounded), symmetric to #429's `minor_compaction_max_files`.
- Minor and major are fenced to **disjoint file-size ranges** via DuckLake's `min_file_size`/`max_file_size` parameters (major now passes `min_file_size := minor_compaction_max_size`), so when both are enabled they run **concurrently with no lock** — validated at startup (`major_compaction_max_size` must exceed `minor_compaction_max_size` when both enabled, else refuse to start).
- `minor_compaction.connection_settings` / `major_compaction.connection_settings` — raw SQL run on that cycle's connection right after opening it, reusing the existing `ConnectionPool.getConnection(String[] sqls)` mechanism. Housekeeping shares major's connection settings.
- Removed `lastMajorAttempt` tracking entirely — no longer needed once major has a real independent cadence.
- Switched compaction scheduling from `scheduleWithFixedDelay` to `scheduleAtFixedRate`: a cycle taking time `T` now waits `frequency - T` before the next one starts (due at `frequency, 2*frequency, ...` from the start), instead of a full `frequency` after completion regardless of `T`.

## Test plan
- [x] `./mvnw test -pl dazzleduck-sql-ducklake-compactor -am` — 25 tests pass across `CompactionConfigOverrideTest`, `CompactionIntegrationTest`, `CompactionOutcomeTest`, including new coverage for enabled-flag combinations, the max-size validation (both directions), connection-settings parsing/application against a real catalog, and independent minor/major failure attribution.


tanejagagan:
Self-review caught two real bugs in the initial version, now fixed:

1. **Thread-pool undersizing**: `compactionScheduler` was still sized to one thread per database, but minor and major are now two independent recurring tasks per database. With one database, that was a 1-thread pool — minor and major would have serialized on it, defeating the entire point of this PR. Now sized per enabled task per database.
2. **Cross-cycle file-count attribution**: the before/after file-count delta feeding `ducklake.files.compacted` was whole-catalog, unscoped. Since minor and major can now genuinely run concurrently, one's merge could get attributed to the other's counter (or double-counted). Fixed by scoping the delta query to each cycle's own file-size range — the same bounds its merge call already uses.

Also fixed: `getStats()`'s next-execution estimate now tracks minor/major's own last-run times separately instead of one shared timestamp, and removed some small duplication (shared SQL-builder for minor/major's merge calls, using `ConnectionPool.getConnection(List)` instead of manual array conversion at each call site).

All 25 tests still green after the fixes.

PR #433, opened 2026-09-17
feat(ingestion): add PartitionedIngestionQueue for hash-sharded parallel writes

tanejagagan:
## Summary
- Adds `PartitionedIngestionQueue`, a `ParquetIngestionQueue` subclass that fans a single flushed bucket out across N concurrent `COPY` shards, routing rows by `hash(partition_column) % parallel_writers = shard`. The accepting side (batching, backpressure, producer-id dedup) is unchanged — parallelism is applied only at flush time.
- Since each shard is its own independent DuckLake transaction, true cross-shard atomicity isn't available. To still guarantee "no partial commit" for the common failure mode, every shard's `COPY` runs to completion before any shard's post-ingestion task runs; a failure during the (much narrower) commit phase after some shards already committed is logged loudly as a partial-commit anomaly rather than hidden.
- `partition_column` / `parallel_writers` are configurable through both the static HOCON `ingestion_queue_table_mapping` and the dynamic SQLite queue registry (with an `ALTER TABLE ADD COLUMN IF NOT EXISTS` migration for registries created before these columns existed).
- Wired into both production `QueueCreator` sites (`DuckDBFlightSqlProducer`, `OtelServiceBase`), which now construct a `PartitionedIngestionQueue` instead of a plain `ParquetIngestionQueue` when a queue's mapping configures partitioning.
- `ParquetIngestionQueue`'s SQL-building/execution methods were widened to `protected` with filter/filename-pattern-aware overloads so the new subclass reuses them instead of duplicating logic; existing single-writer behavior and generated SQL are unchanged (verified — see Test plan).

## Test plan
- [x] Empirically verified against the installed DuckDB 1.5.5 CLI that concurrent `COPY ... FILENAME_PATTERN '{uuid}' ... APPEND` into the same Hive-partitioned directory from separate connections is collision-free (no lost/duplicated rows).
- [x] New `PartitionedIngestionQueueTest`: disjoint/complete hash-bucket routing, `parallelWriters=1` degrades to identical output as plain `ParquetIngestionQueue`, all-shards-fail leaves zero catalog commits, per-shard watermark row computation.
- [x] New `PartitioningMappingTest`: `QueueIdToTableMapping` validation (`parallel_writers > 1` requires `partition_column` and vice versa), HOCON provider parsing, `DuckLakeIngestionHandler` lookup.
- [x] `DynamicQueueRepositoryTest` additions: new columns round-trip through `loadAll`, and a pre-existing (older-schema) SQLite registry is migrated correctly by `init()`.
- [x] `./mvnw test -pl dazzleduck-sql-commons -am` — 527 tests, 0 failures (includes all pre-existing ingestion tests, confirming the `ParquetIngestionQueue` visibility refactor changed no generated SQL for existing callers).
- [x] `./mvnw test -pl dazzleduck-sql-flight,dazzleduck-sql-otel-collector -am` — compiles and passes with the new `QueueCreator` branch in both modules.


dazzleduck-sql-commons/src/main/java/io/dazzleduck/sql/commons/ingestion/DynamicQueueRepository.java
```diff
@@ -69,8 +69,15 @@ public synchronized void init() throws SQLException {
                     input_schema     TEXT,
                     partition_by     TEXT,
                     min_bucket_size  INTEGER,
-                    max_delay_ms     INTEGER
+                    max_delay_ms     INTEGER,
+                    partition_column TEXT,
+                    parallel_writers INTEGER
                 )""".formatted(ATTACHMENT));
+            // Migration for a registry created before partition_column/parallel_writers existed:
```
dazzleduck-web: we do  not need migration for now. Just update the fields for now

tanejagagan:
Pushed a follow-up fix: the all-shards-failed message had a string-concatenation bug (`.formatted()` bound only to the last literal in the concatenation, so `queueId`/failure counts were never substituted — the literal `%s`/`%d` printed instead). Fixed, and every per-shard failure entry now also names its `hash(...) % N` filter alongside its index, so a partial-failure message tells you exactly which rows were affected, not just which shard number.

Also worth flagging for review: on any shard failure today, **all** producers whose batches were combined into that flush get their futures completed exceptionally — not just the producer(s) whose rows happened to hash into the failing shard. This is inherent to two things: `BulkIngestQueue` already combines multiple producers' batches into one bucket before `write()` is called, and routing is per-row (`hash(partition_col) % N`), so a single producer's batch will typically have rows spread across every shard anyway. Isolating failure to just the affected producer(s) is possible (tag rows with their source file via `filename=true`, query which files fed the failed shard, complete unaffected futures before rethrowing) but is a real follow-up scope, not something folded into this PR — happy to take it as a separate issue if wanted.

Full `dazzleduck-sql-commons` suite re-verified after this fix: 527 tests, 0 failures.

tanejagagan:
Pushed the per-producer isolation follow-up discussed above.

**What changed:**
- New `BulkIngestQueue.PartialWriteFailure`: a `write()` implementation that determines only *some* batches in a flushed bucket were affected can now complete every future itself (successful ones normally, affected ones exceptionally) and throw this instead of a plain exception. `BulkIngestQueue.processWriteQueue` recognizes it specially — accounting (`totalWrite*`/`failedWrite*`) and producer-sequence rollback are scoped to exactly the named failed batches, leaving the rest recorded as successfully written. Any other exception still fails the whole bucket, unchanged.
- `PartitionedIngestionQueue` now uses this: when some shards fail (at COPY or at catalog-commit) and others succeed, it re-evaluates just the failed shard(s)' `hash(...) % N` filter against the raw (pre-transformation) input tagged with `filename=true`, to determine exactly which input batches had rows routed to a failed shard. Only those batches' futures fail — batches confined to succeeded shards get their futures completed normally.
- Attribution gracefully degrades to today's whole-bucket-failure behavior when it isn't possible: every shard failing (nothing to disambiguate), or the routing column not existing pre-transformation (e.g. it's derived by the transformation) — logged clearly rather than silently guessed.
- Failure messages now always include the shard index *and* its literal routing filter (`shard 1 [hash("id") % 3 = 1]: ...`), not just the index.

**Known trade-off, documented in code:** rollback for the partial case runs slightly after future completion (both happen on the write thread, but futures complete inside `write()` before the `PartialWriteFailure` is thrown and caught), reopening a narrow, best-effort race — a retry arriving in that microsecond window could be spuriously rejected as out-of-sequence once. Recoverable by the client's normal retry; not a correctness issue, just a marginally weaker guarantee than the whole-bucket path already had.

**Tests added:**
- `PartitionedIngestionQueueTest.testPartialFailureNotifiesOnlyAffectedBatches`: two producers in one flushed bucket, one shard's commit fails — confirms the unaffected producer's future succeeds while the affected one fails, with the error naming the exact shard/filter, and asserts 2 of 3 shards actually committed.
- `BulkIngestQueueTest.testPartialWriteFailureAccountsAndRollsBackOnlyAffectedBatches`: base-class-level test (no DuckDB involved) verifying accounting is split correctly and only the failed producer's sequence rolls back (its retry with the same batch id succeeds; the unaffected producer's sequence was never touched).

Full suite re-verified after this change: `dazzleduck-sql-commons` 529 tests, `dazzleduck-sql-flight` 151, `dazzleduck-sql-otel-collector` 62 — all passing, `BUILD SUCCESS`.

tanejagagan:
Ran a full code review pass on this PR and fixed everything it surfaced:

**1. [High] Silent misattribution when the routing column's name collides with an unrelated raw column.** `attributeAffectedFiles` evaluated `hash(partitionColumn)` on the *raw, pre-transformation* input as a shortcut. If the transformation redefined a column under the same name as `partitionColumn` (e.g. deriving `id` from a `real_key` field, while the raw input also happened to carry an unrelated `id` column), the query found the raw column and computed the wrong value *silently* — no error, just a wrong answer — which could report success for a batch that was actually lost, or fail one that wasn't. Fixed: attribution now runs `constructSourceRelation` (the exact same transformation pipeline the real per-shard write uses) individually per batch, so there's no way for it to resolve a different column than production did. Added `testAttributionUsesTransformedValueNotRawColumnOfSameName`, which fails under the old code and passes under the fix.

**2. [Medium] One malformed row in the dynamic SQLite registry could block reloading every other queue.** `DynamicQueueRepository.loadAll()` had no per-row isolation, so the new `partition_column`/`parallel_writers` cross-field validation throwing on one row aborted the whole method — and since `DynamicIngestionHandler.checkAndReload()` only advances its watermark on success, that one bad row would silently block *all* queues' registry updates on every subsequent poll until fixed. Fixed: an inconsistent pair now degrades just that one queue to single-writer with a logged warning, isolated from the rest. (Static HOCON config keeps failing loud at startup — that's the right behavior there, since an operator sees it immediately before anything runs.)

**3. [Medium] Interrupt cascade in `writeAllShards`.** Catching one shard's `InterruptedException` and immediately re-interrupting the thread made every *subsequent* shard's `Future.get()` in the same loop throw too, misreporting shards that may have genuinely succeeded as failed during a close()-triggered shutdown. Fixed: the interrupt flag is now restored only after every shard has had a chance to report its real outcome.

**4. [Low] `maxProducerIds` leaked a failed producer's id into a successful result.** During a genuine partial split, the merged `IngestionResult` handed to *unaffected* batches' futures was built from the whole bucket's producer-id map, including the producer whose batch actually failed. Fixed: `mergeResults` now scopes `maxProducerIds` to only the batches it actually covers.

**5. [Low/cosmetic] Misleading "attribution unavailable" wording.** Resolved as a side effect of fix #1 — attribution now always resolves definitively per batch, so there's no more separate "unavailable" state to word carefully.

Verified via the full suite: `dazzleduck-sql-commons` 530 tests, `dazzleduck-sql-flight` 151, `dazzleduck-sql-otel-collector` 62 — all passing, `BUILD SUCCESS`.

PR #434, opened 2026-09-18
feat(ingestion): partition a queue into N hash-routed sub-queues

tanejagagan:
## Summary

Adds **per-queue partitioning** to ingestion. A queue configured with `num_partitions > 1` is split into **N independent child queues**, each writing to its own `p<index>` sub-directory of the target path. All children share the queue's table, transformation and DuckLake catalog registration.

- A batch's partition is computed as `hash(partition_expression) % num_partitions` over the raw input rows.
- A batch whose rows all map to a single partition is forwarded to that child.
- A batch whose rows span **more than one** partition is **rejected** with an error (HTTP 400 / Flight/otel `INVALID_ARGUMENT`) and its staged input file deleted — the producer is responsible for pre-partitioning its data.

`partition_expression` is any SQL expression over the input row, e.g. `source_ip` or `substr(source_ip, 1, 10)`.

## Design

`PartitionedIngestionQueue extends ParquetIngestionQueue`, so it drops into every place a queue is handed around. HTTP, Flight and the OTLP collector all resolve queues through `getOrCreateQueue(...) -> queue.add(batch)`, so **all three ingestion paths** get partition routing with no protocol-specific changes (HTTP inherits it via the Flight producer). Metric accessors aggregate across the children.

Config (`num_partitions`, `partition_expression`) is per-queue in `QueueIdToTableMapping`, read from **HOCON** and the **SQLite dynamic registry** (with a column migration for existing registries). Invalid config (`num_partitions > 1` without an expression) fails at load.

## Files
- **commons:** `ConfigConstants` (keys); `QueueIdToTableMapping` (+`num_partitions`/`partition_expression`, validation); `IngestionHandler`/`DuckLakeIngestionHandler` (accessors); `DuckLakeIngestionTaskFactoryProvider` (HOCON); `DynamicQueueRepository` (SQLite columns + migration); new `PartitionedIngestionQueue`
- **flight:** `DuckDBFlightSqlProducer.createQueue` builds the partitioned queue when configured
- **otel:** `OtelServiceBase` creator — same
- **docs:** partitioning examples in the flight + otel `reference.conf`

## Testing
- New `PartitionedIngestionQueueTest`: single-partition write to the correct child sub-dir, distinct-child routing, mixed-batch rejection + input-file cleanup, invalid-config guards.
- HOCON partition-config loading tests.
- Full `dazzleduck-sql-commons` suite green (528 run, 0 failures/errors). Flight/otel/http test-compile clean; their full integration suites (Docker/testcontainers) were not run locally.


tanejagagan:
Follow-up commit `6768bd2f` adds a **per-queue stats dashboard** on top of the partitioning work in this PR (they share the queue plumbing, so they're stacked here rather than in a separate PR):

- **Main server** `/v1/ui` bulk-ingestion table and a **new `/stats` page on the OTLP collector** (its health server), both rendered by a shared `StatsHtml` renderer so columns match.
- **Full per-queue counter set:** rows written; pending bytes + % of `max_pending_write`; data/post-ingest phase ms and avg write ms/bucket; rejections split by reason (**429 backpressure**, **out-of-sequence**, **multi-partition** — the last counted by `PartitionedIngestionQueue` on a rejected mixed batch); freshness (last-write / last-receive age) and last error.
- **Partitioned queues** aggregate children into the top row and render one indented child row per partition (`p0..pN-1`).

Tests: `StatsHtmlTest`, `HealthServerTest` (`/stats` serves HTML), and `PartitionedIngestionQueueTest` (stats aggregation + multi-partition reject counter). Full `dazzleduck-sql-commons` suite green.

Deliberate follow-up: watermark last-commit *values* (last max-timestamp / snapshot id) — they need post-ingestion plumbing and aren't counters.

tanejagagan:
Addressed the scheduler-leak review finding in `46ee253d`:

The Flight path created a fresh non-daemon `Executors.newSingleThreadScheduledExecutor()` **per queue**, never shut down — leaking one thread per queue and **N+1 per partitioned queue** on eviction. `DuckDBFlightSqlProducer` now owns a **single daemon-threaded `bulkIngestFlushScheduler`** shared by every bulk-ingest queue and all partition children, shut down in `close()` — mirroring the OTLP collector, which already shares one flush scheduler. A lingering scheduled trigger for an evicted queue is a no-op (`triggerWriteIfRequired` returns once `terminating`), so sharing is safe. `createQueue` now takes the scheduler as a parameter.

Also updated the `partition_expression` examples to realistic map/struct access (`resource['user_id']`, `claims['user_id']`).

All modules compile; commons partition + stats tests green.

PR #436, opened 2026-09-18
feat(compactor): per-cycle compaction telemetry + /ui dashboard

tanejagagan:
Implements the capture-only telemetry from **#435** (`COMPACTION_TELEMETRY_SPEC.md`). No change to scheduling, cadence, or batch sizing.

## What's captured
One `CompactionRun` per cycle per `(database, tier)`:
- **Identity/timing:** run_id, scheduled/started/ended, intended_delay, actual_gap
- **Work:** band files + **bytes** before/after (a same-round-trip `SUM(file_size_bytes)` added to the `COUNT` the compactor already ran — bytes are the real cost driver), files_retired, groups_requested
- **Duration:** total + merge (see deviation below)
- **Outcome:** `SUCCESS | EMPTY | FAILED`, a classified `failure_class` (COMMIT_TIMEOUT / TRANSACTION_CONFLICT / OUT_OF_MEMORY / SPILL_EXCEEDED / CATALOG_UNAVAILABLE / OTHER), truncated error
- **Resource headroom:** RSS `VmHWM`, temp-dir spill, memory_limit

## Storage & exposure
- **Ring buffer** (`CompactionRunLog`, last 50 per (db,tier)) computing the spec's derived quantities: throughput files/bytes, **arrival rate**, drain rate, p95 duration, **duration headroom** (p95 / commit timeout), saturated, idle ratio.
- **`GET /ui`** — a dependency-free HTML dashboard (`CompactionHtml`), one card per tier with those aggregates, failure-class chips, resource line, and inline-SVG sparklines of cycle duration and files retired.
- **Micrometer** — bytes retired per tier, failures tagged by `class`.

## The one non-capture change (per the spec)
Moved `updateAllTierFileCounts` **off the per-cycle control path** onto its own `file_count_refresh_frequency` schedule (default 30s) with a dedicated connection — roughly halving per-cycle catalog traffic, which is what makes the capture affordable.

## Decisions
- **Merge/commit split not taken:** DuckLake merges *and* commits inside one autocommit `CALL ducklake_merge_adjacent_files(...)`, so there is no separate JDBC commit to time. `duration_commit_ms = -1`; `duration_merge_ms` is the whole CALL. Splitting would require manual commit — a behaviour change the spec's "capture only" scope forbids.
- **JSONL sink deferred** to a follow-up; ring buffer + `/ui` + metrics ship here.
- **groups_merged left null** (spec Q2 unresolved); the CALL's result set is drained.
- Spec Q3 (raising `idle_in_transaction_session_timeout` via the ATTACH options) is a behaviour change and out of scope here.

## Config (new keys, all with safe defaults)
`file_count_refresh_frequency` (30s), `run_history_size` (50), `commit_timeout_limit` (2m, only feeds durationHeadroom).

## Tests
`CompactionRunTest` (failure taxonomy), `CompactionRunLogTest` (derived-quantity formulas, ring eviction), `ResourceSamplerTest`, `CompactionHtmlTest`, updated `CompactionOutcomeTest` — all green. (The Testcontainers `CompactionIntegrationTest` needs Docker/Postgres; run in CI.)


tanejagagan:
Follow-up `d6c4937c`: the telemetry now exports over the **existing OTLP gRPC pipeline** (no new export machinery), and the `/health` JSON carries a per-tier telemetry section.

New Micrometer meters (per db+tier unless noted):
- `groups_requested` and `tier_frequency` (gauges) — control inputs emitted now so making them **dynamic** later is observable
- `drain_rate`, `arrival_rate` (gauges) — deliberately emitted because they can't be reconstructed by rating cumulative counters
- `saturated` (0/1), `spill_peak_bytes` (gauges); `rss_peak_bytes` (one per process — VmHWM is monotonic)
- `cycles_by_outcome` (counter, tagged SUCCESS/EMPTY/FAILED)

Gauge holders are updated once per cycle (no per-scrape recompute). `GET /health` gains a `telemetry` object per tier (latest run + derived aggregates) for programmatic scraping alongside `/ui`.

**No JSONL sink** — confirmed not required. Metric set is otherwise as recommended.

dazzleduck-sql-ducklake-compactor/src/main/java/io/dazzleduck/sql/compaction/CompactionRun.java
```diff
@@ -0,0 +1,111 @@
...
+        long actualGapMs,
+        Long bandFilesBefore,
+        Long bandFilesAfter,
+        Long filesRetired,
+        Long bandBytesBefore,
+        Long bandBytesAfter,
+        long groupsRequested,
+        Long groupsMerged,
+        long durationTotalMs,
+        long durationMergeMs,
```
dazzleduck-web, line 41: Do we capture both durationMergeMs and durationCommitMs. We need to remove them if they are not captured properly

tanejagagan: Good catch — fixed in dbf95530. `durationCommitMs` was always -1 (DuckLake merges + commits inside one autocommit CALL, so there's no separate commit to time without switching to manual commit — a behaviour change the capture-only scope forbids), so I removed the field entirely from `CompactionRun` and `TierCompactor.MergeOutcome` rather than carry a permanent sentinel. `durationMergeMs` (the whole CALL) is captured properly and kept.

Note: `groupsMerged` is also still always null (spec Q2 — the CALL's result shape isn't relied on yet). I left it as a nullable 'unknown' rather than a fake value since the spec defines the field, but happy to drop it too by the same principle if you'd prefer.

tanejagagan:
Addressed the review + did a self-review pass (`dbf95530`):

- **Removed `durationCommitMs`** (always -1; not capturable without a behaviour change) from `CompactionRun` and `TierCompactor.MergeOutcome`. `durationMergeMs` — captured, kept.
- **Added `HealthServerTest`** covering the new exposure the review pass flagged as untested: `/health` carries per-tier telemetry (latest run + derived aggregates) when runs exist and `null` before any run; `/ui` serves the dashboard.
- Review pass found no functional bugs. One thing I left, flagged on the inline thread: `groupsMerged` is still always null (spec Q2) — kept as a nullable "unknown" since the spec defines the field, but I'll drop it too if you'd rather not carry an uncaptured field.

Compactor unit tests: HealthServerTest 3, CompactionRunTest 5, CompactionRunLogTest 6, ResourceSamplerTest 6, CompactionHtmlTest 4, CompactionOutcomeTest 10 — all green. (Testcontainers `CompactionIntegrationTest` runs in CI.)

dazzleduck-sql-ducklake-compactor/src/main/java/io/dazzleduck/sql/compaction/CompactionRun.java
tanejagagan: Renamed + populated in daf71f26:

- `groupsRequested` → `maxCompactedFiles` (it's exactly the tier's max_compacted_files), `groupsMerged` → `compactedFiles`. The gauge `ducklake.compaction.groups_requested` is now `ducklake.compaction.max_compacted_files`, and /health reports both maxCompactedFiles and compactedFiles.
- `compactedFiles` is now populated from the CALL's result set when the merge is bounded (max_compacted_files > 0). Since the exact result shape of ducklake_merge_adjacent_files is still unconfirmed (Q2), it's provisional — captured as the returned row count — and DuckDbTierCompactor logs the real shape once at INFO so we can confirm/correct it against a live catalog. Null for unbounded merges.

tanejagagan: Confirmed the shape by running an actual merge (DuckLake on a DuckDB-file catalog, no Postgres needed — DuckDB v1.5.5):

```
CALL ducklake_merge_adjacent_files('lake', min_file_size := 0, max_file_size := 1000000000, max_compacted_files := 10);
-- schema_name | table_name | files_processed | files_created
-- main        | t          | 4               | 1
```

So it returns one row per compacted table with files_processed / files_created, whenever a merge happens (not gated on max_compacted_files). Updated `compactedFiles` to be SUM(files_processed) (the input files compacted), null when no rows. Added `DuckDbTierCompactorMergeTest` that runs a real merge over a DuckDB-file catalog and asserts compactedFiles == 4, then null on the no-op second cycle — runs in the normal unit suite, no Testcontainers. (6f347922)

tanejagagan: Added the second field (3bcafbcc). Since the merge returns both counts, the record now mirrors DuckLake's columns exactly:

- **filesProcessed** = SUM(files_processed) — input files compacted away (this is what compactedFiles held)
- **filesCreated** = SUM(files_created) — new output files produced

Both null when nothing merged. Carried through MergeOutcome, CompactionRun, and /health telemetry; DuckDbTierCompactorMergeTest asserts 4 processed / 1 created on a real merge.

PR #437, opened 2026-09-19
perf: disable DuckDB's external file cache in compactor and collector

gtaneja-sentrinox:
## Summary

Disable DuckDB's `enable_external_file_cache` in the compactor and the OTLP collector. Both read each external file exactly once and then discard it, so the cache can never hit — it only accumulates memory.

- **Compactor** — a merge reads its source files once and then retires them.
- **Collector** — the ingest path stages each batch to an Arrow file, reads it once via `read_arrow([...])`, then deletes it (`ParquetIngestionQueue`).

## Measured

Benchmark compactor, before and after, from a fresh start each time:

| | before | after |
|---|---|---|
| minor tier RSS | **14.8 GiB** after ~44 min | **~0.7 GiB**, flat |
| major tier RSS | **8.1 GiB** (at its 8 GB limit) | **~1.1 GiB**, flat |
| host memory | 23 GB of 30 GB used | ~2 GB |

Minor was climbing steadily from 0.5 GiB at restart to 14.8 GiB in 44 minutes. After the change it oscillates in a 0.6–0.8 GiB band — a working set rather than a ramp.

**Throughput is unchanged.** Major cycles ran **17.9 s mean** after versus **17–22 s** before, and neither tier's band depth or failure count moved. The cache was buying nothing on this workload.

Note it is **not bounded by `memory_limit`** — which is why a 12 GB limit did not contain a 14.8 GiB RSS, and why this looked like a leak for two days before we found it.

## Not the footer caches

Worth recording, since that was the first hypothesis: the metadata caches were already off by default and were never involved.

```
parquet_metadata_cache       false
enable_http_metadata_cache   false
enable_object_cache          false   [PLACEHOLDER] legacy, does nothing
enable_external_file_cache   true    <- the actual cause, caches file DATA
```

## Placement

`enable_external_file_cache` is **GLOBAL**-scoped, so it is applied per DuckDB instance:

- `RawConnections` (compactor) — before the startup script and before a tier's `connection_settings`, so an operator can still override it.
- `Main` (collector) — once on the singleton, which is the only DuckDB instance in that process.

**Deliberately not set inside `ConnectionPool`**, which the flight and query paths share. There the same cache serves repeated reads of the same files and does earn its keep; making it a global default would be a regression for query workloads.

## Test plan

- [x] `./mvnw compile -pl dazzleduck-sql-ducklake-compactor,dazzleduck-sql-otel-collector -am` — clean.
- [x] Compactor change validated in production-shape deployment (figures above), both tiers, over ~20 min each with cycle duration and failure counts compared.
- [ ] **Collector change is not separately measured.** It is justified by the identical read-once-then-delete access pattern in `ParquetIngestionQueue` rather than by a before/after measurement, because validating it costs an ingest restart. Reviewers may want that measured before merge.


dazzleduck-sql-otel-collector/src/main/java/io/dazzleduck/sql/otel/collector/Main.java
```diff
@@ -40,6 +40,14 @@ public static void main(String[] argv) throws Exception {
         // Run the startup script before calling toProperties() so that extensions like
         // ducklake and arrow are loaded before DuckLakeIngestionHandler queries DuckLake metadata.
         CollectorConfig config = new CollectorConfig(args.configFile);
+
+        // The ingest path stages each batch to an Arrow file, reads it exactly once via
+        // read_arrow([...]) and then deletes it, so DuckDB's external-file cache (an in-memory LRU
+        // over external file DATA) can never hit and only costs memory. GLOBAL-scoped, so one SET on
+        // the singleton covers this process; deliberately not set inside ConnectionPool itself,
+        // which query paths share and where repeated reads make the cache worthwhile.
+        ConnectionPool.executeOnSingleton("SET enable_external_file_cache = false");
```
gtaneja-sentrinox: This is very sneaky. Should come from better place perhaps configuration

gtaneja-sentrinox:
Fair — that was hidden behaviour an operator could neither discover nor override. Moved it into configuration in 4cf19cf.

The default now lives in `reference.conf`, next to `min_bucket_size` and `max_delay_ms` where operators already look:

```hocon
ingestion {
    min_bucket_size     = 1048576
    max_delay_ms        = 5000
    parquet_compression = snappy

    # Raw SQL applied to the ingestion DuckDB instance at startup.
    #
    # enable_external_file_cache is an in-memory LRU over external file DATA. The ingest
    # path stages each batch to an Arrow file, reads it exactly once via read_arrow([...])
    # and then deletes it, so the cache can never hit and only costs memory. It is also not
    # bounded by memory_limit. Set to [] to restore DuckDB's default, or add your own SETs.
    connection_settings = ["SET enable_external_file_cache = false"]
}
```

`Main` now iterates the configured list instead of executing a literal, so `[]` restores DuckDB's default and other `SET` statements can be added without a code change.

Two deliberate choices worth flagging:

- **Naming mirrors the compactor's per-tier `connection_settings`** — same idea, same reason, so the two services read consistently.
- **Kept out of the shared `IngestionConfig` record**, which the flight module also constructs. This is a collector-side concern, and widening that record would push it onto an unrelated consumer.

The compactor half of this PR is unchanged. Its equivalent (`RawConnections.WORKLOAD_DEFAULTS`) is applied *before* each tier's `connection_settings`, so a tier can already override it — but if you'd prefer that surfaced in config too rather than defaulted in code, say so and I'll move it the same way.

gtaneja-sentrinox:
Done for the compactor too, in e1cf1a7 — you're right that it already had the config surface for this.

`WORKLOAD_DEFAULTS` is gone entirely; **`RawConnections.java` is now byte-identical to `main`**, so this PR no longer touches compactor Java at all. The setting moved into the shipped `application.conf`, alongside the other per-connection settings:

```hocon
compaction_tiers = [
  { name = "minor"
    connection_settings = ["SET enable_external_file_cache = false"]
    # e.g. add "SET memory_limit='2GB'", "SET threads=2"
  }
  { name = "major"
    connection_settings = ["SET enable_external_file_cache = false"]
    # e.g. add "SET memory_limit='8GB'", "SET threads=4"
  }
]

housekeeping_connection_settings = ["SET enable_external_file_cache = false"]
```

The rationale is documented where `connection_settings` is already described, including one gotcha worth flagging explicitly: **HOCON replaces a list wholesale rather than merging it**, so an operator who supplies their own `compaction_tiers` block has to repeat this setting to keep it. That's the real cost of preferring config over a code default here, and it seemed better to document it than to hide it.

Net effect of the PR now:

| module | change |
|---|---|
| compactor | `application.conf` only — shipped defaults + docs |
| collector | `connection_settings` key, accessor, `Main` loop, `reference.conf` default |
| common | one `ConfigConstants` key |

No hardcoded DuckDB settings remain in either service.

PR #441, opened 2026-09-21
feat(compactor): bake the patched DuckLake extension into native + jib images

tanejagagan:
## Summary

DuckLake's compaction-conflict check operates at table granularity: two transactions compacting the same table always conflict, even over completely disjoint files. That blocks this module's N-tier design (#431) from ever running tiers concurrently against one table. [`duckdb/ducklake#1453`](https://github.com/duckdb/ducklake/pull/1453) fixes it (not yet merged upstream) by escalating the check to file granularity.

This PR bakes a patched DuckLake extension build into `dazzleduck-sql-ducklake-compactor`'s images instead of the stock one:

- The patch is backported onto `d8a1881e` — the exact commit DuckDB `v1.5.5` pins via its own `ducklake.cmake` — in [`dazzleduck-web/ducklake`](https://github.com/dazzleduck-web/ducklake), branch `backport/1453-file-level-compaction-conflict`. Verified ABI-matched to this repo's `duckdb_jdbc:1.5.5.1` directly via `pragma_version()`/`duckdb_extensions()`, not just inferred from the pin file.
- That repo's `build-and-publish-extension.yml` (manually triggered) builds `linux_amd64`/`linux_arm64` via DuckDB's own official `extension-ci-tools` reusable workflow (manylinux-style environment, avoiding a glibc mismatch against our runtime images), then auto-publishes both binaries as release assets — see [`v1.5.5-dazzleduck.1`](https://github.com/dazzleduck-web/ducklake/releases/tag/v1.5.5-dazzleduck.1).
- **`Dockerfile.native`**: `curl`s the matching-arch asset into DuckDB's local extension cache path before the existing pre-fetch install step, instead of `INSTALL ducklake`.
- **`pom.xml`**: an `exec-maven-plugin` execution downloads the arch-specific asset ahead of `jib:build`, baked in via `extraDirectories`.
- **`RawConnections`**: sets `allow_unsigned_extensions` as a connection property (not a `SET` statement — DuckDB rejects changing it once the database instance is running, caught by an actual test failure while building this) since the build isn't signed with DuckDB Labs' key.
- **`DuckLakePatchVersionTest`** (new): fails the build if `duckdb.version` and `ducklake.patch.built.for.duckdb.version` ever drift apart, so a future DuckDB bump can't silently ship images with an ABI-mismatched or stale patched extension.
- **`DUCKLAKE_PATCH.md`** (new): what the patch is, where the build/publish pipeline lives, and the exact steps to update or retire it.

Scoped to the compactor only — no other module loads DuckLake or changes here.

## Test plan
- [x] `./mvnw clean install` (full reactor) — green
- [x] `./mvnw test -pl dazzleduck-sql-ducklake-compactor` end-to-end (no bypass) — real download from the release, all 55 tests pass
- [x] Extension build+publish pipeline itself green for both archs
- [ ] `native-ci.yml` (triggers automatically on this PR) — builds + smoke-tests the native image for both archs


gtaneja-sentrinox:
## Review

Reviewed at `cb93b00e`. One release blocker, then some build-hygiene items. The critical finding was reproduced in a `linux/amd64` container against the real release asset, not just read off the diff.

### 1. HIGH — both shipped images crash at startup on `LOAD ducklake`

`RawConnections.java:44` sets `allow_unsigned_extensions=true` as a connection property, and the comment there is right that `SET` won't work once the instance exists. But that covers only the *raw* compaction and housekeeping connections. `Main.java:27` runs the same startup script on the **shared commons singleton first** — deliberately, so `TableConfigProvider` can read its table out of the ATTACHed catalog:

```java
String startupScript = readStartupScript(rawConfig);
if (startupScript != null) {
    ConnectionPool.executeOnSingleton(startupScript);   // <-- no allow_unsigned_extensions
}
```

That singleton's properties come from `ConnectionPool.loadProperties()`, which reads exactly one classpath resource, `duckdb.properties`. **No such file exists anywhere in the repo** (`git ls-tree -r` finds none), so the singleton runs with `allow_unsigned_extensions=false`.

Meanwhile the image pre-places the unsigned patched binary at the cache path DuckDB resolves `LOAD` from — the pom comment at line 243 says exactly that, and names "the compactor's startup script's `LOAD ducklake`". Every real startup script does contain it, because it is what makes `ATTACH 'ducklake:...'` work: see `example/startup/ducklake.sql` (baked into the jib base image) and `CompactionIntegrationTest.java:67`.

`ConnectionPool.executeOnSingleton` throws `RuntimeSqlException` on failure, so this propagates straight out of `main` — a hard startup crash, not a degraded mode.

Reproduced in a container with `duckdb_jdbc-1.5.5.1` and the real asset:

| case | result |
|---|---|
| `LOAD` only, no allow_unsigned | **FAIL** — signature is either missing or invalid |
| `LOAD` only, allow_unsigned | OK, `extension_version = fb555937` (the patched build) |
| `INSTALL`+`LOAD`, no allow_unsigned, **network up** | **FAIL** — same error |

The third row is the one worth dwelling on: having network access does not rescue it. `INSTALL` is a no-op because the file already exists, so it never falls back to the stock signed extension.

Two reasons CI is green regardless:

- Tests run on a host with a clean extension cache, where `INSTALL` fetches the *stock signed* extension.
- `native-ci.yml`'s smoke test starts the container with no startup script, so `startupScript == null` and line 27 is skipped entirely.

**Fix:** ship `dazzleduck-sql-ducklake-compactor/src/main/resources/duckdb.properties` containing `allow_unsigned_extensions=true`. That is the only hook `ConnectionPool` offers, and per the existing comment the setting cannot be applied via `SET` after the instance is created.

One caveat on that fix: `duckdb.properties` is resolved from the classloader, so it is process-global. Fine for an executable module like this one, but it would leak into the server's own `ConnectionPool` if this jar ever became a dependency of it. Worth a comment in the file saying so.

### 2. MEDIUM — the drift guard misses the one pin that silently breaks the image

`DuckLakePatchVersionTest` asserts `duckdb.version == ducklake.patch.built.for.duckdb.version`, but not `ducklake.extension.cache.version` (`pom.xml:44`).

On a DuckDB bump to, say, `1.6.0.0`, the test forces you to update `built.for` and in practice `ducklake.patch.release` — but leave the cache version at `v1.5.5` and both `Dockerfile.native:45` and jib's `into` place the binary under `extensions/v1.5.5/linux_<arch>/` while DuckDB looks in `extensions/v1.6.0/...`. `curl` succeeds, the test passes, the image builds green, and the container then either fails at runtime or silently network-installs the **stock** extension — losing the patch, which is precisely the "silently ship a stale patched extension" outcome this test is documented to prevent.

The value is derivable (`v` plus the first three components of `duckdb.version`), so assert it in the same test.

### 3. MEDIUM — 36 MB download on every build, not just image builds

The `download-patched-ducklake-extension` execution is bound to `generate-resources` (`pom.xml:196`), which is part of the default lifecycle. The comment explains this was to sequence ahead of the unbound `jib:build` goal, but the cost is that all of these re-fetch the 36 MB asset with no caching or `skipIfExists`:

- `./mvnw test -pl dazzleduck-sql-ducklake-compactor`
- the documented `./mvnw clean package install -DskipTests`
- `ci.yml`'s `./mvnw -B -ntp clean verify`
- `release.yml`'s "Install reactor" step

`curl -fsSL` exits non-zero on an unreachable or rate-limited github.com, or a deleted release, failing the whole reactor build for an artifact only `jib:build` needs.

Separately, the inner `./mvnw -Pnative ... package` at `Dockerfile.native:26` also triggers this execution, and it uses `${jib.architecture}` (default `amd64`) rather than `TARGETARCH` — so an arm64 native build downloads a 36 MB amd64 binary it never uses, and would hard-fail if only the arm64 asset were published.

Suggest moving it into a profile, or making it a no-op when the target file already exists.

### 4. LOW — release images bake in both architectures' extensions

`release.yml` runs `./mvnw $JIB_ARGS package jib:build -pl dazzleduck-sql-ducklake-compactor -Djib.architecture=$arch` for `amd64` then `arm64` with no `clean` between iterations, so by the second pass `target/ducklake-extension/` holds both `linux_amd64/` and `linux_arm64/`. `extraDirectories.from` copies the directory wholesale (`pom.xml:250`), so the arm64 image ships an unused 36 MB amd64 binary — and vice versa for anyone building in the other order.

Functionally harmless, since DuckDB resolves by platform name, but it is ~36 MB of dead weight per published image. Pointing `from` at `target/ducklake-extension/linux_${jib.architecture}` with `into` at `/root/.duckdb/extensions/${ducklake.extension.cache.version}/linux_${jib.architecture}` fixes it.

### 5. LOW — the doc describes something this PR establishes is impossible

`DUCKLAKE_PATCH.md:56` ("How the images consume it") says the verification program runs:

```sql
SET allow_unsigned_extensions = true;
LOAD ducklake;
```

But `Dockerfile.native:59-61` uses a JDBC connection property precisely because DuckDB rejects `SET` on that option once the instance is running — as the code comment and the PR description both state. A maintainer following this doc during a future patch refresh will write the form that cannot work. The "Runtime enablement" table row already has it right; line 56 should match.

### Verified as fine

`microdnf install -y zlib-devel curl` succeeds on `ghcr.io/graalvm/native-image-community:21` — curl is already present and simply gets upgraded — so that hunk is good. The cache-path placement and the `allow_unsigned_extensions` mechanism are both correct in themselves; finding 1 is only about the one connection that misses the property.

---

Finding 1 is a release blocker as it stands: the images build, pass CI, and fail to start with any real startup script. Happy to send a patch for it plus the version-drift assertion if useful.

tanejagagan:
Thanks for the thorough review — findings 2 through 5 fixed as suggested:

- **#2**: `DuckLakePatchVersionTest` now also asserts `ducklake.extension.cache.version` matches `duckdb.version`'s first three components.
- **#3**: the extension download is skipped by default (`ducklake.extension.download.skip=true`); `release.yml`, `scripts/docker-publish.sh`, this module's README, and `.claude/commands/publish-docker.md` now pass `-Dducklake.extension.download.skip=false` where the jib image is actually built. This also fixes the native-build wrong-arch download you flagged, since `Dockerfile.native`'s inner `-Pnative` build never passes that flag.
- **#4**: `extraDirectories` now scopes to `linux_${jib.architecture}` only.
- **#5**: doc fixed to match the actual (connection-property) mechanism.

**#1 (HIGH)**: implemented the suggested `duckdb.properties` fix — but I want to report honestly that I could not reproduce the claimed crash. I ran three separate tests against the real built image (`jib:dockerBuild`, no push), each starting the singleton with a real startup script:

1. `LOAD ducklake;` alone, no `allow_unsigned_extensions` anywhere — succeeded.
2. `INSTALL ducklake; LOAD ducklake;`, no `allow_unsigned_extensions` — succeeded. Confirmed via `sha256sum` inside the container that `INSTALL` really is a no-op (file hash unchanged, still our patched build) — it isn't silently falling back to the stock signed extension either.
3. Full `ATTACH 'ducklake:/tmp/test.ducklake' AS testdb (...)` with `databases=[testdb]` — succeeded; the service actually started ("Compaction service started for 1 database(s)..."). The only error that followed was `RawConnections` independently re-attaching the same file-based catalog concurrently from a second DuckDB instance — a known, unrelated limitation (file-based DuckLake catalogs don't support concurrent attach; `RawConnections`' own test file already documents this), nothing to do with signing.

I don't have a theory for why `LOAD`-by-name from the local extension cache wouldn't enforce the signature check the same way `INSTALL` (or an explicit-path `LOAD`) does, but that's what three independent runs against the shipped artifact show. I kept the `duckdb.properties` fix anyway — it's harmless, and gives the singleton the same defensive posture as `RawConnections` regardless of whether today's DuckDB build strictly requires it. If you have a specific repro (exact startup script, DuckDB build, or a different code path) that shows the crash, I'd like to see it so I can re-verify — happy to be wrong here, but wanted to report what I actually observed rather than just take the fix on faith.

Separately, and unrelated to this review: while testing I initially used `jib:build` instead of `jib:dockerBuild` for local verification and accidentally pushed a `0.2.20-SNAPSHOT` image to `dazzleduck/ducklake-compactor:latest-amd64` on Docker Hub. Caught it immediately and restored it (`docker buildx imagetools create` back to the real `0.2.19-amd64` digest) — confirmed the actual public `:latest` multi-arch manifest was never affected (it's a separate manifest-list that only updates when the release workflow's manifest step runs). Flagging for transparency.

